PR #9, opened 2020-12-02
Add recording support

apotterri:
Add appmap.Recording, a context manager that records code executed in
its block.


dustinbyrne (approved):
LGTM! 👍

tests/test_recording.py
```diff
@@ -0,0 +1,80 @@
...
+    import appmap
+    from appmap._implementation import generation
+    r = appmap.Recording()
+    with r:
+        from src import Src
+        Src().instance_method()
+
+    generated_appmap = orjson.loads(generation.dump(r))
+    for event in generated_appmap['events']:
+        for k, v in event.items():
```
virajkanwade, line 76:
```suggestion
    for event in generated_appmap['events']:
        event.update((k, os.path.basename(v)) for k, v in event.items() if k == 'path')        
```

OR

```suggestion
    for event in generated_appmap['events']:
        if 'path' in event:
            event['path'] = os.path.basename(event['path'])
```

tests/test_recording.py
```diff
@@ -0,0 +1,80 @@
...
+    # import here to use the environment variables set above
+    import appmap
+    from appmap._implementation import generation
+    r = appmap.Recording()
+    with r:
+        from src import Src
+        Src().instance_method()
+
+    generated_appmap = orjson.loads(generation.dump(r))
+    for event in generated_appmap['events']:
```
virajkanwade, line 75:
```suggestion
    for event in generated_appmap['events']:
        event.update((k, os.path.basename(v)) for k, v in event.items() if k == 'path')        
```

OR

```suggestion
    for event in generated_appmap['events']:
        if 'path' in event:
            event['path'] = os.path.basename(event['path'])
```

tests/test_recording.py
```diff
@@ -0,0 +1,80 @@
...
+        for k, v in event.items():
+            if k == 'path':
+                event[k] = os.path.basename(v)
```
virajkanwade, line 78:
```suggestion
    for event in generated_appmap['events']:
        event.update((k, os.path.basename(v)) for k, v in event.items() if k == 'path')        
```

OR

```suggestion
    for event in generated_appmap['events']:
        if 'path' in event:
            event['path'] = os.path.basename(event['path'])
```

appmap/_implementation/configuration.py
```diff
@@ -0,0 +1,147 @@
...
+    return os.getenv("APPMAP", "false") == "true"
+
+def split_method_name(method):
+    qualname = method.__qualname__
+    if '.' in qualname:
+        class_name, func_name = qualname.rsplit('.', 1)
+    else:
+        class_name = ''
+        func_name = qualname
+    return (class_name, func_name)
```
virajkanwade, line 29:
```suggestion
    """ Split method __qualname__ into class_name, func_name """
    return ([''] + method.__qualname__.rsplit('.', 1))[-2:]
```

appmap/_implementation/configuration.py
```diff
@@ -0,0 +1,147 @@
...
+            return wrap(method)
+        return self.next_filter.call(method)
+
+
+
+class BuiltinFilter(Filter):
+
+    @classmethod
+    def load_config(cls):
+        if not enabled:
```
virajkanwade, line 132:
```suggestion
        if not enabled():
```

tests/test_recording.py
apotterri: I'm now passing an `object_hook` to `json.loads` that takes care of this. (It's using `in` the way you suggested here.)

appmap/_implementation/configuration.py
apotterri: `enabled()` has moved into `appmap._implementation.env`, and that's now what's being used here.

apotterri:
@virajkanwade This is ready for a final look, any additional comments would be welcomed.

I've left many of the changes in separate commits, just to make them easier to see. Once you've had a chance to take a look, I'll squash them into one or two, add a CHANGELOG entry, and get things merged.

This is showing a build failure, but looks like there was a network hiccup that messed things up. There's a green build here of the latest commit here: https://travis-ci.com/github/applandinc/appmap-python/jobs/460816419

virajkanwade:
@apotterri quick comment: appmap/test/data/test_class.py rename this file. Pytest will look for file names starting with test_* and assume it to pytest file.

apotterri:
Oops, thanks for this. I changed the names.

virajkanwade:
@apotterri just curious. tests folder was renamed to test?

appmap/test/test_recording.py
```diff
@@ -0,0 +1,64 @@
...
+    )
+
+
+@pytest.mark.datafiles(
+    os.path.join(FIXTURE_DIR, 'appmap.yml'),
+    os.path.join(FIXTURE_DIR, 'example_class.py'),
+    os.path.join(FIXTURE_DIR, 'expected.appmap.json')
+    )
+def test_recording_works(monkeypatch, datafiles):
+    with open(datafiles / 'expected.appmap.json') as f:
```
virajkanwade, line 22:
```suggestion
    with open(os.path.join(datafiles, 'expected.appmap.json')) as f:
```

appmap/test/test_recording.py
```diff
@@ -0,0 +1,64 @@
...
+        expected_appmap = json.load(f)
+
+    # Setting these outside the definition of expected_appmap makes it
+    # easier to update when the expected appmap changes
+    py_impl = platform.python_implementation()
+    py_version = platform.python_version()
+    expected_appmap['metadata']['language']['engine'] = py_impl
+    expected_appmap['metadata']['language']['version'] = py_version
+
+
```
virajkanwade, line 32:
```suggestion
```

appmap/test/test_recording.py
```diff
@@ -0,0 +1,64 @@
...
+    sys.path.append(str(datafiles))
+
+    monkeypatch.setenv("APPMAP", "true")
+    monkeypatch.setenv("APPMAP_CONFIG", str(datafiles / 'appmap.yml'))
```
virajkanwade, line 36:
```suggestion
    monkeypatch.setenv("APPMAP_CONFIG", str(os.path.join(datafiles, 'appmap.yml')))
```

appmap/test/test_recording.py
```diff
@@ -0,0 +1,64 @@
...
+    monkeypatch.setenv("APPMAP_LOG_LEVEL", "debug")
+
+    import appmap
+    reload(appmap._implementation.recording)  # pylint: disable=protected-access
```
virajkanwade, line 40: why?

appmap/_implementation/generation.py
```diff
@@ -0,0 +1,117 @@
...
+
+from .event import Event, serialize_event
+
+
+class ClassMapSet(MutableMapping):
+    def __init__(self):
+        self.dict = dict()
+
+    def __delitem__(self, k):
+        del self[k]
```
virajkanwade, line 15: Should this be: 
```suggestion
        del self.dict[k]
```

appmap/_implementation/generation.py
```diff
@@ -0,0 +1,117 @@
+"""Generate an AppMap"""
+import json
+from collections.abc import MutableMapping
+
+from . import metadata
+
+from .event import Event, serialize_event
+
+
+class ClassMapSet(MutableMapping):
```
virajkanwade, line 10: Doesn't seem to be doing anything specifically different from `dict`. Subclassing `dict` might give similar behavior.

appmap/_implementation/generation.py
```diff
@@ -0,0 +1,117 @@
...
+    def __hash__(self):
+        return hash(self.key)
+
+
+class PackageEntry(ClassMapEntry):
+    def __init__(self, name):
+        super().__init__(name, name, 'package')
+        self.children = ClassMapSet()
+
+    __hash__ = ClassMapEntry.__hash__
```
virajkanwade, line 48: @apotterri won't `__hash__` be inherited?

appmap/_implementation/generation.py
```diff
@@ -0,0 +1,117 @@
...
+        self.dict[k] = v
+
+    def __iter__(self):
+        return self.dict.__iter__()
+
+    def __len__(self):
+        return len(self.dict)
+
+
+class ClassMapEntry:
```
virajkanwade, line 30: Also check https://docs.python.org/3/library/dataclasses.html

appmap/_implementation/generation.py
```diff
@@ -0,0 +1,117 @@
...
+    def __init__(self, e, loc):
+        super().__init__(loc, e.method_id, 'function')
+        self.path = e.path
+        self.lineno = e.lineno
+        self.static = e.static
+
+    __hash__ = ClassMapEntry.__hash__
+
+
+def asdict(s):
```
virajkanwade, line 69: @apotterri why a separate function?

apotterri:
> @apotterri just curious. tests folder was renamed to test?

Yes. I like the aesthetics of having the tests as part of the application code, so I followed the layout suggested in the [pytest doc](https://docs.pytest.org/en/stable/goodpractices.html#tests-as-part-of-application-code) (and also found in various other places). Practicality trumps aesthetics, though, so if there's an advantage to having it at the root of the repo, I can move it back.

appmap/test/test_recording.py
apotterri: I wrote it this way, because the [pytest-datafiles](https://pypi.org/project/pytest-datafiles/) doc uses the `/` operator. Happy to change it, but what's the advantage to using `os.path.join`?

appmap/test/test_recording.py
apotterri: Reloading `recording` was necessary to reinitialize it to use the environment variable settings. In thinking about how to answer your question, I realized I didn't really like having most of the initialization done as a side-effect of loading the package.

So, I create an `initialize` function for several of the modules, and called each explicitly. Those changes are in e025a0d.

appmap/_implementation/generation.py
apotterri: I guess this was based on a misunderstanding of when to subclass `dict` vs when to subclass `MutableMapping`. I'll switch to `dict.

appmap/_implementation/generation.py
apotterri: No, per https://docs.python.org/3/reference/datamodel.html#object.__hash__ , which says:
> If a class that overrides __eq__() needs to retain the implementation of __hash__() from a parent class, the interpreter must be told this explicitly by setting __hash__ = <ParentClass>.__hash__.

appmap/test/test_recording.py
virajkanwade: it is useful only when on Windows which uses \ instead of /. But I guess they way Windows is moving towards linux base, might not be an issue long term.

appmap/_implementation/generation.py
virajkanwade: based on the current implementation it looks like we are just renaming `dict` to ClassMapSet. If you think in future, there could be more custom code in the getter setter or some other internal implementation can leave it as is.

appmap/_implementation/generation.py
virajkanwade: from what I have understood, if the subclass overrides `__eq__`, only then its `__hash__` is set to None.

```>>> class ClassMapEntry:
...     def __init__(self, key, name, entry_type):
...         self.key = key
...         self.name = name
...         self.type = entry_type
...     def __eq__(self, other):
...         return self.key == other.key
...     def __hash__(self):
...         return hash(self.key)
...
>>> class PackageEntry(ClassMapEntry):
...     def __init__(self, name):
...         super().__init__(name, name, 'package')
...         self.children = ClassMapSet()
...
>>> PackageEntry.__hash__
<function ClassMapEntry.__hash__ at 0x10acc6e50>
>>> class ClassEntry(ClassMapEntry):
...     def __init__(self, name):
...         super().__init__(name, name, 'class')
...         self.children = ClassMapSet()
...     def __eq__(self, other):
...         pass
...
>>> ClassEntry.__hash__
>>> PackageEntry.__hash__
<function ClassMapEntry.__hash__ at 0x10acc6e50>```

Comment (virajkanwade) [reply appmap/_implementation/generation.py:48]: anyways, doesn't hurt either ways

**Comment by virajkanwade** [comment]
> > @apotterri just curious. tests folder was renamed to test?
> 
> Yes. I like the aesthetics of having the tests as part of the application code, so I followed the layout suggested in the [pytest doc](https://docs.pytest.org/en/stable/goodpractices.html#tests-as-part-of-application-code) (and also found in various other places). Practicality trumps aesthetics, though, so if there's an advantage to having it at the root of the repo, I can move it back.

Nah. question wasn't about the location. You had renamed `test` to `tests` previously and in this PR, you have renamed it back to `test`. Was curious about the naming, not the location.

File: appmap/test/test_recording.py (hunk 6 above)
Comment (apotterri) [reply appmap/test/test_recording.py:21]: Rereading the doc, `datafiles` is a `py.path`. So, `datafiles / 'expected.appmap.json'` uses `py.path.__div__` to construct a path with the correct separator. (If they were strings, trying to use the division operator would raise an exception, right?)

But, having said that, I see that my copying and pasting from the pytest-datafiles doc has left us with a mix of `/` and `os.path.join`, with more uses of `os.path.join` than `/`. I'll switch to `os.path.join` to make them consistent.

File: appmap/_implementation/generation.py (hunk 10 above)
Comment (apotterri) [reply appmap/_implementation/generation.py:15]: Yes. Fixed.

File: appmap/_implementation/generation.py (hunk 12 above)
Comment (apotterri) [reply appmap/_implementation/generation.py:48]: I confirmed that you're correct, subclasses of `ClassMapEntry` use its `__hash__` method. I've taken the assignment out of the subclasses.

I've taken your other suggestion, though, and switched to dataclasses, eliminating the need for any of this.

File: appmap/_implementation/generation.py (hunk 13 above)
Comment (apotterri) [reply appmap/_implementation/generation.py:30]: Done.

File: appmap/_implementation/generation.py (hunk 14 above)
Comment (apotterri) [reply appmap/_implementation/generation.py:69]: This was a holdover from an implementation using `orjson`, which required a separate function. I've moved this into `AppMapEncoder`.

File: appmap/_implementation/generation.py (hunk 11 above)
Comment (apotterri) [reply appmap/_implementation/generation.py:10]: You're right about the intention here, which just amounts to renaming `dict`. However, when I switched to `class ClassMapSet(dict)`,  the tests failed. So, I think I'll leave this as-is, for now, and switch to subclassing `dict` in a future round of changes.

File: appmap/_implementation/generation.py (hunk 13 above)
Comment (apotterri) [reply appmap/_implementation/generation.py:30]: Well, actually, I tried to switch to dataclasses, but I didn't like the end result. Unless you object, I'll leave these as regular classes.

File: appmap/_implementation/py_version_check.py [hunk 15]
```diff
@@ -0,0 +1,16 @@
...
+class AppMapPyVerException(Exception):
+    pass
+
+
+def check_py_version():
+    cur_py_version = [int(x) for x in platform.python_version_tuple()]
+
+    min_py_version = [int(x) for x in str(MIN_PY_VERSION).split('.')]
+
+    if cur_py_version[0] < min_py_version[0] or cur_py_version[1] < min_py_version[1]:
```
virajkanwade, line 15:
```suggestion
    MIN_PY_VERSION = (3, 8)
    if sys.version_info < MIN_PY_VERSION:
```

came across this snippet

appmap/_implementation/configuration.py
```diff
@@ -0,0 +1,175 @@
...
+
+        logger.debug('  undecided')
+        return self.next_filter.filter(class_)
+
+    def wrap(self, fn, isstatic):
+        if self.excluded(fn):
+            return fn
+        if self.included(fn):
+            if getattr(fn, '_appmap_wrapped', None) is None:
+                return wrap(fn, isstatic)
```
virajkanwade, line 143: Should we rename global `def wrap` to make it less confusing since this class method is also the same name.

apotterri:
Closing in favor of #11.

PR #12, opened 2020-12-13
Recording 20201212

virajkanwade:
@apotterri major code refactor.

If this looks good, you could skip pr #11


apotterri:
@virajkanwade Thanks very much for this. Sorry I was slow to pick up your last set of changes, I've been busy fixing some major implementation problems.

All of these changes look good to me, except maybe the refactoring in 337dfef. I think we need to keep the separation of concerns that this commit removes. `generation.py` is currently pretty sparse, but there's going to be a lot more functionality in it soon (with my current changes, in fact). Having a file that specifies the interface for filters make sense, but it doesn't seem like all the disparate filters should get grouped together. I also think it will be clearer to keep the filter stack external from Recording.

virajkanwade:
This PR is based on top of #11 . So we can surely keep the refactor code aside. I will create single PRs for small improvements which could be incorporated in the current code.

PR #14, opened 2020-12-13
Singleton Recorder class

virajkanwade:


apotterri (commented):
Are there other changes that need to be made as a result of this? Shouldn't the `recorder` variable get replaced with references to `Recorder()`?

virajkanwade:
@apotterri that would be optional. I had done the changes in my big refactor commit. Making the class singleton would ensure that if somebody accidentally calls Recorder() it would still return the  same instance.

PR #20, opened 2021-01-05
Protect against recursion when inspecting a call event's receiver

apotterri:
Also, simplify the classes used to generate the classmap.


appmap/_implementation/configuration.py
```diff
@@ -52,7 +69,11 @@ def in_set(name, which):
 
 
 def function_in_set(fn, which):
-    class_name, fn_name = utils.split_function_name(fn)
+    class_name, fn_name = split_function_name(fn)
+    if class_name is None:
+        # Is this really the right thing to do?
```
virajkanwade, line 74: @apotterri what is the aim here?

appmap/_implementation/generation.py
```diff
@@ -1,67 +1,48 @@
...
-        self.name = name
-        self.type = entry_type
-
-    def __eq__(self, other):
-        return self.key == other.key
-
-    def __hash__(self):
-        return hash(self.key)
+    name: str
+    type: str
```
virajkanwade, line 16: @apotterri `type` is a system variable

appmap/_implementation/configuration.py
apotterri: I'm not sure how bare functions (i.e. functions that aren't in a class) are supposed to be represented in an AppMap. Until I figure it out, this just skips instrumenting them.

appmap/_implementation/generation.py
apotterri: How much of a problem is this? Python isn't currently objecting to it, and the AppMap attribute needs to be named `type`. I'm sure it can be changed if absolutely necessary....

virajkanwade: nah, scratch that. its an attribute not variable, so it should be fine.

appmap/_implementation/event.py
```diff
@@ -29,9 +31,16 @@ def next_thread_id(cls):
...
     def get_thread_id(cls):
         tls = utils.appmap_tls()
-        if 'thread_id' not in tls:
-            tls['thread_id'] = cls.next_thread_id()
-        return tls['thread_id']
+        try:
+            thread_id = tls['thread_id']
+        except KeyError:
+            thread_id = tls['thread_id'] = cls.next_thread_id()
+        return thread_id
```
virajkanwade: try this
`return tls.setdefault('thread_id', cls.next_thread_id())`

virajkanwade: `>>> d = {}
>>> d.setdefault('a', 2)
2
>>> d
{'a': 2}
>>> d['a'] = 3
>>> d.setdefault('a', 2)
3`

appmap/_implementation/event.py
```diff
@@ -151,4 +160,8 @@ def initialize():
     # If True, allow calls to str() and repr() to raise RuntimeErrors
     # (e.g. Recursion Error). This makes it possible to test
     # instrumentation's recursion avoidance.
-    _EventIds.id = 1
+    try:
+        del utils.appmap_tls()['thread_id']
+    except KeyError:
+        pass
+    _EventIds.reset()
```
virajkanwade, line 167: `suggestion
utils.appmap_tls.pop('thread_id', None)
`

appmap/_implementation/event.py
apotterri: I don't think we can do that. `cls.next_thread_id()` should only be called if `thread_id` doesn't exist in `tls`.

appmap/_implementation/event.py
```diff
@@ -151,4 +160,8 @@ def initialize():
     # If True, allow calls to str() and repr() to raise RuntimeErrors
     # (e.g. Recursion Error). This makes it possible to test
     # instrumentation's recursion avoidance.
-    _EventIds.id = 1
+    try:
+        del utils.appmap_tls()['thread_id']
+    except KeyError:
+        pass
```
virajkanwade, line 166: suggestion
`utils.appmap_tls.pop('thread_id', None)`

appmap/_implementation/event.py
virajkanwade: ahh right. It will populate next_thread_id even if value is found. my bad.

virajkanwade: few alternate oneliners

```
>>> def foo():
...     print('foo')
...     return 2
...
```

```
>>> d = {}
>>> d['a'] if 'a' in d else d.setdefault('a', foo())
foo
2
>>> d['a'] if 'a' in d else d.setdefault('a', foo())
2
```

```
>>> d = {}
>>> d.get('a') or d.setdefault('a', foo())
foo
2
>>> d.get('a') or d.setdefault('a', foo())
2
```

apotterri: I believe performance will be important here, and the `try...except` idiom seems to be used quite a bit. It seems to me that your second suggestion has all the benefits of the `try...except` construct, but reads better. I'll switch to that.

appmap/_implementation/configuration.py
apotterri: I've updated the comment to indicate what's going on here.

apotterri:
@virajkanwade This feels like it's done, so I'm going to merge it. If there are other things you'd like to see changed, though, feel free to add more comments to this PR, or open another one. Thanks for your help!

PR #22, opened 2021-01-07
No recursion 20201223

virajkanwade:
fix for tests randomly failing

if test_events.py is executed before other tests, it sets the _EventIds._next_thread_id and does not reset it which causes the next test case to fail as thread_id is now 6.


apotterri:
Thanks for this. I pushed 05cf611 yesterday, which also fixes the test failures. I'll make sure the flake8 formatting changes get applied, too.

PR #31, opened 2021-01-27
Add pytest integration

apotterri:
Add pytest integration. Fix some issues that were uncovered along the way.


virajkanwade:
@apotterri 
```
## [Unreleased]
### Added
- WIP - pytest integration

### Changed
- Simplify (and fix) logging configuration
  The code that configured logging was overly complex, and also buggy.
  These changes simplify it, keep duplicate messages from being emitted,
  and properly allow per-module configuration of log level.
- Don't use inspect.isclass
  When testing to see if an object is a class, use type() instead of
  inpect.isclass. See the comment on
  appmap._implementation.recording.is_class for details.
  Also, make sure ConfigFilter.wrap doesn't call the next filter for a
  function that it finds has already been wrapped.

### Fixed
- Fix classmap `function` entries
  Entries of type `function` in the classmap must have an attribute called
  "location", not "path" and "lineno" attributes. Now they do.


```

apotterri:
Yeah, thanks, I'll definitely make sure the CHANGELOG gets updated by the time the PR is ready to go.

Did you generate these entries manually, or do you have something automatic to do it?

virajkanwade:
Manually as you need to categorize them

apotterri:
> Manually as you need to categorize them

Yeah, although it seems like the comments could be extracted from the commits automatically, at least to get things started.

apotterri:
@virajkanwade Will you take a look at this when you have a chance, please? Thanks!

CHANGELOG.md
```diff
@@ -7,7 +7,39 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ### Added
 - [#17] Add elapsed attribute to ReturnEvent
-- [#8] add git to metadata
+- [#8] Add git to metadata
+- pytest integration
+- Support `APPMAP_OUTPUT_DIR` environment variable. If set, specifies the root directory
+  for writing AppMaps. If not set, defaults to `tmp/appmap`.
+
+### Changed
+- Use repr() instead of str() for receiver
```
virajkanwade, line 16: @apotterri Changelog should be a summary of changes since last release. I believe these 2 entries could be skipped

appmap/_implementation/configuration.py
```diff
@@ -6,19 +6,59 @@
...
+            cls._instance._initialized = not env.enabled()
+
+        return cls._instance
+
+    def __init__(self):
+        if self.__getattribute__('_initialized'):  # keep pylint happy
+            return
+
+        config_file = os.getenv("APPMAP_CONFIG", "appmap.yml")
+        with open(config_file) as file:
```
virajkanwade, line 42: @apotterri is error handling required if appmap.yml does not exist?

CHANGELOG.md
apotterri: Hmmm. I think all the entries here are still unreleased. Which two entries did you mean?

appmap/_implementation/configuration.py
apotterri: We only try to load appmap.yml if `APPMAP=true` is in the environment. If it doesn't exist, you get a stack trace that ends with 
```
FileNotFoundError: [Errno 2] No such file or directory: 'appmap.yml'
```
which seems fine to me (and matches the behavior of appmap-ruby).

CHANGELOG.md
virajkanwade: @apotterri sorry, was confused. Please ignore.

appmap/pytest.py
```diff
@@ -0,0 +1,127 @@
...
+    logger.debug('pytest_pyfunc_call, metadata %s', repr(metadata))
+    fname = item.filename + '.appmap.json'
+    path = os.path.join(session.appmap_path, fname)
+    def write_recording(r):
+        with open(path, 'wb') as appmap:
+            appmap.write(generation.dump(r, metadata))
+        logging.info('wrote recording %s', path)
+
+    logging.info('starting recording %s', path)
+    r = recording.Recording(exit_hook=write_recording)
```
virajkanwade, line 125: I ran it against a simple pytest, the events and classMap was empty. Is that expected?

```{
    "version": "1.4",
    "metadata": {
        "language": {
            "name": "python",
            "engine": "CPython",
            "version": "3.9.0"
        },
        "client": {
            "name": "appmap",
            "url": "https://github.com/applandinc/appmap-python"
        },
        "app": "Capitalize",
        "frameworks": [
            {
                "name": "pytest",
                "version": "6.2.1"
            }
        ],
        "recorder": {
            "name": "pytest"
        },
        "name": "raises exception on non string arguments",
        "feature": "Raises exception on non string arguments"
    },
    "events": [],
    "classMap": []
}```

Comment (apotterri) [reply appmap/pytest.py:125]: Maybe. It really depends on what the test does, and what's in appmap.yml. In general, it's probably more likely to happen with a simple test than with a test that does more.

Your question points out something that's missing in this PR, though: it doesn't have any tests showing the pytest integration works. I'll add some.

File: README.md [hunk 4]
```diff
@@ -23,7 +37,9 @@ appropriate test framework support into your tests.
 `test_*` function in the subclasses.
 
 ### pytest
-`import appmap.pytest`. Instruments and records `test_*` functions.
+`import appmap.pytest`. Instruments and records `test_*` functions. Resulting AppMaps are
```
virajkanwade, line 40: It didn't work for me with `import appmap.pytest` in the test file.

I was able to get it to work by placing `pytest_plugins = ['appmap.pytest']` in the conftest.py

Will investigate if I get some time

virajkanwade: https://docs.pytest.org/en/stable/writing_plugins.html#conftest-py-local-per-directory-plugins

pytest loads plugin modules at tool startup in the following way:

1. by scanning the command line for the -p no:name option and blocking that plugin from being loaded (even builtin plugins can be blocked this way). This happens before normal command-line parsing.
2. by loading all builtin plugins.
3. by scanning the command line for the -p name option and loading the specified plugin. This happens before normal command-line parsing.
4. by loading all plugins registered through setuptools entry points.
5. by loading all plugins specified through the PYTEST_PLUGINS environment variable.
6. by loading all conftest.py files as inferred by the command line invocation:
    - if no test paths are specified, use the current dir as a test path
    - if exists, load conftest.py and test*/conftest.py relative to the directory part of the first test path. After the conftest.py file is loaded, load all plugins specified in its pytest_plugins variable if present.
  
    Note that pytest does not find conftest.py files in deeper nested sub directories at tool startup. It is usually a good idea to keep your conftest.py file in the top level test or project root directory.

7. by recursively loading all plugins specified by the pytest_plugins variable in conftest.py files.

virajkanwade: Three options to include appmap in pytest I can think of:

1. Place `pytest_plugins = ['appmap.pytest']` in _conftest.py_
2. Run with `APPMAP=true ./.venv/bin/pytest -p appmap.pytest -s`
3. `PYTEST_PLUGINS=appmap.pytest APPMAP=true ./.venv/bin/pytest -s`

virajkanwade: pytest-django might be something to look at

apotterri: Yeah, the pytest integration is done by providing a pytest plugin. I'll fix the doc.

Once it's available through PyPI, you'll be able to `pip install appmap` into a project that uses pytest and the appmap plugin will get used.

To test it locally, in a shell that's had a virtual environment activated, you can run `poetry install` in the root of your `appmap-python` repo:

```
ajp@Alans-MacBook-Pro appmap-python % pyenv activate
pyenv-virtualenv: version `flaskbb' is already activated
(flaskbb) params_20210202 *
ajp@Alans-MacBook-Pro appmap-python % pwd
/Users/ajp/src/applandinc/appmap-python
(flaskbb) params_20210202 *
ajp@Alans-MacBook-Pro appmap-python % poetry install
Installing dependencies from lock file

No dependencies to install or update

Installing the current project: appmap (0.0.0)
(flaskbb) params_20210202 *
ajp@Alans-MacBook-Pro appmap-python % pip show appmap
Name: appmap
Version: 0.0.0
Summary: Create AppMap files by recording a Python application.
Home-page: None
Author: Alan Potter
Author-email: alan@app.land
License: None
Location: /Users/ajp/.pyenv/versions/3.9.0/envs/flaskbb/lib/python3.9/site-packages
Requires: orjson, inflection, PyYAML
Required-by:
(flaskbb) params_20210202 *
ajp@Alans-MacBook-Pro appmap-python % pytest --collect-only | head -20
============================= test session starts ==============================
platform darwin -- Python 3.9.0, pytest-6.2.0, py-1.10.0, pluggy-0.13.1
Using --randomly-seed=239152410
rootdir: /Users/ajp/src/applandinc/appmap-python, configfile: pytest.ini
plugins: appmap-0.0.0, pyfakefs-4.3.2, datafiles-2.0, cov-2.11.1, randomly-3.5.0, xdist-2.2.0, forked-1.3.0, mock-3.5.1
collected 20 items
```

apotterri:
@virajkanwade This is ready for another look.

Thanks!

appmap/pytest.py
```diff
@@ -0,0 +1,129 @@
...
+        return fname.removesuffix('_')
+
+    @property
+    def metadata(self):
+        ret = {}
+        if self.has_feature_group():
+            ret['feature_group'] = self.feature_group
+            ret['recording'] = {
+                'defined_class': self.defined_class,
+                'method_id': self.method_id
```
dividedmind, line 72: Won't these be present even without a feature group?

appmap/test/appmap_test_base.py
```diff
@@ -1,6 +1,64 @@
...
+    def normalize_appmap(self, generated_appmap):
+        # Normalize paths
+        object_id = 1
+
+        def normalize(dct):
+            nonlocal object_id
+            if 'elapsed' in dct:
+                assert isinstance(dct['elapsed'], float)
+                dct['elapsed'] = 0.0
+            if 'git' in dct:
```
dividedmind, line 22: Would it make sense to remove `git` instead, or set it to `None`? Perhaps after verifying the field types. (Same with `elapsed`.)

appmap/test/data/pytest/pytestdebug.log
```diff
@@ -0,0 +1,564 @@
+versions pytest-6.2.2, py-1.10.0, python-3.9.1.final.0
```
dividedmind, line 1: Should this have been committed?

appmap/test/test_pytest.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class TestPytest(AppMapTestBase):
+    def compose_exec(self, args, test_dir, tmpdir):
+        cwd = test_dir
+        env = {
+            "PATH": os.getenv("PATH"),
+            "APPMAP_OUTPUT_DIR": tmpdir
+        }
+        return subprocess.run(['docker-compose'] + args,
```
dividedmind, line 17: Would it be possible to not use docker here? Using docker here is quite unexpected. I don't have an appropriate python available in my main environment and tried to run the tests in docker. This failed utterly. (Of course it's possible to make it work, but it's fiddly and brittle.)

It seems to work to pretty much just call `poetry run pytest` in the correct directory instead of docker-compose. A separate dockerfile and script can be provided to run the whole thing in docker for convenience.

appmap/test/test_configuration.py
```diff
@@ -19,3 +25,108 @@ def test_is_disabled_when_false(self, monkeypatch):
...
+    def test_fn_prefix_doesnt_match(self, datafiles, mocker, monkeypatch):
+        monkeypatch.setenv("APPMAP", "true")
+        monkeypatch.setenv("APPMAP_CONFIG",
+                           os.path.join(str(datafiles), 'appmap.yml'))
+        monkeypatch.setenv("APPMAP_LOG_LEVEL", 'debug')
+
+        f = mocker.Mock()
+        f.__module__ = 'package1_prefix'
+        f.__qualname__ = 'cls.func'
+        fltr = ConfigFilter(NullFilter())
```
dividedmind, line 131: Most of this is duplicated a lot which makes the code hard to follow. It seems most of it can be DRYed up.

dividedmind (changes requested):
Looks like it's working! Had a couple minor comments, but I'd say it's almost ready.

appmap/_implementation/configuration.py
```diff
@@ -6,19 +6,63 @@
...
 from . import event
 from .recording import Recorder, Filter
-from .utils import appmap_tls, split_function_name
+from .utils import appmap_tls, split_function_name, fqname
 
 logger = logging.getLogger(__name__)
 
 
+class Config:
+    """ Singleton Config class """
```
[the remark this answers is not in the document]
apotterri: Yeah, perhaps. How would you suggest providing the default instance to its users?

poetry.lock
```diff
@@ -65,6 +65,14 @@ mccabe = ">=0.6.0,<0.7.0"
 pycodestyle = ">=2.6.0a1,<2.7.0"
 pyflakes = ">=2.2.0,<2.3.0"
 
+[[package]]
```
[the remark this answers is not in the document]
apotterri: Yup, that's a good idea. I'll take it out.

appmap/_implementation/configuration.py
```diff
@@ -79,15 +128,18 @@ def function_in_set(fn, which):
         # fn isn't in a class, ignore it.
         return False
 
-    class_name += '.'
-    logger.debug(('class_name %s'
+    ret = (class_in_set(class_name, which)
+           or name_in_set(f'{class_name}.{fn_name}', which))
```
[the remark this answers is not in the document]
apotterri: You're right, it won't. Changing it would make it consistent with Ruby, sort of, and not consistent with Java, which uses `.` for instance and static methods. 

Also, `appmap-ruby` doesn't allow you to references to methods by name in `appmap.yml`. So it's inconsistent from that perspective as well.

appmap/_implementation/configuration.py
```diff
@@ -114,62 +166,68 @@ class ConfigFilter(Filter):
     full names will match.  (This is achieved in the code by appending
     `.` to the name retrieved from the config file.)
```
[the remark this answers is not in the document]
apotterri: Nope. I'll fix the comment.

appmap/pytest.py
apotterri: Oops, yes, thanks. Looks like it got auto-indented, I'll fix it.

appmap/test/appmap_test_base.py
apotterri: @virajkanwade You added this filtering. What do you think about @dividedmind's suggestion?

appmap/test/test_configuration.py
apotterri: Yeah, I hadn't really come to grips with pytest fixtures yet when I wrote this. I understand them better now, and they seem like a good fit for cleaning this up.

appmap/test/test_pytest.py
apotterri: The most important thing here is to ensure that installing `appmap-python` into a project that uses `pytest` configures everything correctly. Using docker is an easy way to do that. Both `appmap-ruby` and `appmap-java` have some docker-based testing for basically the same purpose. 

There is a Dockerfile and an entrypoint script as part of this PR. It should be easy enough to get rid of the use docker-compose, but I think we should keep using docker for the test.

appmap/test/appmap_test_base.py
virajkanwade: @apotterri @dividedmind Had initially thought of removing it, but then this atleast validates if the appropriate keys are being set.

appmap/test/test_pytest.py
virajkanwade: @apotterri have you had a chance to look at this: https://docs.pytest.org/en/stable/writing_plugins.html#testing-plugins

appmap/_implementation/configuration.py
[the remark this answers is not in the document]
virajkanwade: Being more Python than Ruby, to me personally, # syntax initially seemed pretty foreign. It is very unlikely for the `appmap.yml` to be shared between a Ruby and Python project, so I, personally, don't see the need for making it consistent across technologies.

appmap/test/data/pytest/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM python:3.9
```
virajkanwade, line 1:
```suggestion
FROM python:3.9-alpine
```

Size difference: 332.72 MB vs 16.44MB

virajkanwade: Altough might need to check dependency compatibility

virajkanwade: Did a quick check. No longer applicable. 
`pip install poetry` has a dependency on `cryptography` which recently added dependency on `rust` which itself is around 392 MB.

https://github.com/pyca/cryptography/issues/5771

virajkanwade: @apotterri I am seeing this error with `python:3.9`

```
=========================================================================================================== ERRORS ===========================================================================================================
_______________________________________________________________________________________________ ERROR collecting test session ________________________________________________________________________________________________
/usr/local/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1030: in _gcd_import
    ???
<frozen importlib._bootstrap>:1007: in _find_and_load
    ???
<frozen importlib._bootstrap>:986: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:680: in _load_unlocked
    ???
../../../_implementation/recording.py:227: in wrapped_exec_module
    exec_module(*args, **kwargs)
../../../_implementation/recording.py:227: in wrapped_exec_module
    exec_module(*args, **kwargs)
/usr/local/lib/python3.9/site-packages/_pytest/assertion/rewrite.py:170: in exec_module
    exec(co, module.__dict__)
tmp/pip-install-53_gxics/cryptography_5ab623b84cd5460694d6723754acd560/tests/conftest.py:10: in <module>
    from .utils import check_backend_support
../../../_implementation/recording.py:227: in wrapped_exec_module
    exec_module(*args, **kwargs)
tmp/pip-install-53_gxics/cryptography_5ab623b84cd5460694d6723754acd560/tests/utils.py:18: in <module>
    import cryptography_vectors
E   ModuleNotFoundError: No module named 'cryptography_vectors'
```

virajkanwade: `pip install cryptography_vectors`

appmap/test/data/pytest/Dockerfile
```diff
@@ -0,0 +1,7 @@
...
+
+RUN pip install poetry
```
virajkanwade, line 3:
```suggestion
RUN pip install poetry cryptography_vectors
```

appmap/test/data/pytest/pytestdebug.log
apotterri: Nope, I'll remove it.

appmap/test/data/pytest/Dockerfile
apotterri: Ugh. I'm losing patience with poetry. It's getting in the way more than it's being helpful. I've been considering switch to just using pip, this might be the final straw.

@dividedmind @virajkanwade What do you guys think of switching to pip?

apotterri: Well, looking around for alternatives, maybe poetry is actually the best choice. it won't be required for users `appmap-python`, so the size increase from requiring rust problably isn't such a big deal.

appmap/test/test_pytest.py
apotterri: Just took a look. It's very specific to pytest, of course. I still feel like a docker-based approach gives us the easiest way to get a clean environment, and will be applicable to test all the different integrations.

appmap/_implementation/configuration.py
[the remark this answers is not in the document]
dividedmind: I'd say, a class method `default` that returns a singleton instance, and whenever config is used, call that to set default but allow overriding (by overriding a default argument or setting a property).

appmap/test/test_pytest.py
dividedmind: The problem with using docker is it makes it quite difficult to run it inside docker... at the very least the docker-based tests should be optional and easily skipped. (It has always annoyed me in `appmap-ruby`, too.)

appmap/test/data/pytest/Dockerfile
dividedmind: I don't think it's a good idea to switch to alpine. The base image is smaller but you end up having to download and compile a lot, so it makes it actually worse. (I know, I went down this rabbit hole.) See https://pythonspeed.com/articles/alpine-docker-python/ for a good write-up on the subject.

It might be worth it if you're creating some specific, limited, user-downloadable, slimmed down image. But it's not what we're doing here.

appmap/test/data/pytest/Dockerfile
dividedmind: This isn't necessary with the full image AFAICT.

appmap/test/test_pytest.py
dividedmind: I believe [tox](https://tox.readthedocs.io/) is a tool that could help setting up the environment to test it all sans docker.

appmap/test/data/pytest/Dockerfile
virajkanwade: @dividedmind  Yes, alpine, in this case turns out to be a bigger overhead. As mentioned above I gave it a try and ended up downloading almost 600-700 MB packages.

appmap/test/data/pytest/Dockerfile
virajkanwade: @dividedmind ran into this issue with `python:3.9` image.

dividedmind: Weird, works for me. Oh well, I suppose it can't hurt, right?

apotterri: I just pulled the latest version of the `python:3.9` image. `pip install poetry` ran fine in it, without needing to specify `cryptography_vectors`.

@virajkanwade Can you pull the image and see if you're able to reproduce this problem, please? If you can, it woul dbe good to figure out why.

appmap/pytest.py
apotterri: Turns out this wasn't a bug, just a poorly named predicate.

There will only be a `defined_class` and `method_id` for this item if the test function is contained in class, which is what the `has_feature_group` predicate tested.

I've changed the name of the predicate so the code will be more clear.

appmap/test/test_pytest.py
apotterri: I looked at tox and at pytest's support for testing plugins. The latter looked like an easier way to get up and running quickly, so I went with that.

appmap/test/data/pytest/Dockerfile
virajkanwade: I believe you have removed docker dependency.

Altough 2 things:
1. `removeprefix` and `removesuffix` functions are 3.9 feature. So we should update the min py version check to `(3, 9) `
2. For some reason, for me, the test fails as the generated appmap has `classMap` empty and `events` empty.

apotterri:
@dividedmind @virajkanwade This is ready for a final look.

I've fixed all the things that were actually broken/wrong. I've also removed docker-based testing, in favor of pytest's support for testing plugins.

If you're ok with it, I'd like to defer the remaining issues raised here till future PRs.

appmap/test/data/pytest/Dockerfile
apotterri: Sorry, I added https://github.com/applandinc/appmap-python/pull/31#issuecomment-779471319 before I'd seen this.

1. I'll fix the version check.

2. Yes, I removed the dependency on Docker. Re the empty AppMap, make sure that the version of pytest.ini you've got has these lines in it.

```
# running in a subprocess ensures that environment variables are set
# correctly and no classes are loaded.
addopts = --runpytest subprocess
```

If it doesn't, please retry with the latest commit. If it does, let me know and I'll try to figure out why.

appmap/test/test_pytest.py
```diff
@@ -0,0 +1,29 @@
...
+            'tmp', 'appmap', 'pytest', 'test_hello_world.appmap.json'
+        )
+        with open(appmap_json) as appmap:
+            generated_appmap = self.normalize_appmap(appmap.read())
+
+        data_dir = os.path.join(pytestconfig.rootpath,
+                                'appmap', 'test', 'data', 'pytest')
+        with open(os.path.join(data_dir, 'expected.appmap.json')) as f:
+            expected_appmap = json.load(f)
+
```
virajkanwade, line 28:
```suggestion
        # Setting these outside the definition of expected_appmap makes it
        # easier to update when the expected appmap changes
        py_impl = platform.python_implementation()
        py_version = platform.python_version()
        expected_appmap['metadata']['language']['engine'] = py_impl
        expected_appmap['metadata']['language']['version'] = py_version

```

appmap/test/test_pytest.py
```diff
@@ -0,0 +1,29 @@
+import os
+import os.path
+
```
virajkanwade, line 3:
```suggestion
import platform

```

appmap/test/test_pytest.py
virajkanwade: I have 3.9.0 so it failed the test.

virajkanwade:
> @dividedmind @virajkanwade This is ready for a final look.
> 
> I've fixed all the things that were actually broken/wrong. I've also removed docker-based testing, in favor of pytest's support for testing plugins.
> 
> If you're ok with it, I'd like to defer the remaining issues raised here till future PRs.

Hi @apotterri Added one suggestion. Rest seems fine to me.

dividedmind (approved):
Thanks for the changes, especially removing the docker dependency! Looks good, I'm okay with merging as soon as you convince travis ;)

appmap/_implementation/configuration.py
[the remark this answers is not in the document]
apotterri: Created an issue for  this: https://github.com/applandinc/appmap-python/issues/33

appmap/test/appmap_test_base.py
apotterri: Done.

appmap/test/test_configuration.py
apotterri: I added issue https://github.com/applandinc/appmap-python/issues/34 for this.

appmap/test/test_pytest.py
apotterri: The new normalization of generated AppMaps incorporates this suggestion.

appmap/test/data/pytest/Dockerfile
ryanmarshall: I have it working with `FROM python:3.9.1-slim-buster` with very minimal installs

PR #37, opened 2021-02-17
Capture SQL queries in django apps

### LINKED ISSUE #26: SQL queries should be captured
When a database client makes a SQL query, a `call` event with a `sql_query` attribute should be generated.

dividedmind:
Fixes #26.

### DIFF
--- a/appmap/_implementation/generation.py
+++ b/appmap/_implementation/generation.py
@@ -42,19 +42,25 @@ class FuncEntry(ClassMapEntry):
 def classmap(recording):
     ret = ClassMapDict()
     for e in recording.events:
-        if e.event != 'call':
-            continue
-        packages, class_ = e.defined_class.rsplit('.', 1)
-        children = ret
-        for p in packages.split('.'):
-            entry = children.setdefault(p, PackageEntry(p))
+        try:
+            if e.event != 'call':
+                continue
+            packages, class_ = e.defined_class.rsplit('.', 1)
+            children = ret
+            for p in packages.split('.'):
+                entry = children.setdefault(p, PackageEntry(p))
+                children = entry.children
+
+            entry = children.setdefault(class_, ClassEntry(class_))
             children = entry.children
 
-        entry = children.setdefault(class_, ClassEntry(class_))
-        children = entry.children
-
-        loc = f'{e.path}:{e.lineno}'
-        children.setdefault(loc, FuncEntry(e))
+            loc = f'{e.path}:{e.lineno}'
+            children.setdefault(loc, FuncEntry(e))
+        except AttributeError:
+            # Event might not have a defined_class attribute;
+            # SQL events for example are calls without it.
+            # Ignore them when building the class map.
+            continue
 
     return ret
 
--- a/appmap/django.py
+++ b/appmap/django.py
@@ -1,0 +1,46 @@
+from appmap._implementation.event import SqlEvent, ReturnEvent
+from appmap._implementation.recording import Recorder
+
+from django.db.backends.signals import connection_created
+from django.dispatch import receiver
+
+import time
+
+
+class ExecuteWrapper:
+    def __init__(self):
+        self.recorder = Recorder()
+
+    def __call__(self, execute, sql, params, many, context):
+        start = time.monotonic()
+        try:
+            return execute(sql, params, many, context)
+        finally:
+            if self.recorder.enabled:
+                stop = time.monotonic()
+                duration = stop - start
+                # Note this logic is based on django.db.backends.utils.CursorDebugWrapper.
+                if many:
+                    # Sometimes the same query is executed with different parameter sets.
+                    # Instead of substituting them all, just say how many times it was run.
+                    try:
+                        times = len(params)
+                    except TypeError:
+                        times = '?'
+                    sql = f'{times} times {sql}'
+                else:
+                    db = context['connection']
+                    cursor = context['cursor']
+                    sql = db.ops.last_executed_query(cursor, sql, params)
+                call_event = SqlEvent(sql)
+                self.recorder.add_event(call_event)
+                return_event = ReturnEvent(parent_id=call_event.id, elapsed=duration)
+                self.recorder.add_event(return_event)
+
+
+@receiver(connection_created)
+def connected(sender, connection, **_):
+    # pylint: disable=unused-argument
+    wrappers = connection.execute_wrappers
+    if not any(isinstance(x, ExecuteWrapper) for x in wrappers):
+        wrappers.append(ExecuteWrapper())
--- a/appmap/pytest.py
+++ b/appmap/pytest.py
@@ -12,6 +12,13 @@ from appmap._implementation.utils import fqname
 
 logger = logging.getLogger(__name__)
 
+try:
+    # Make sure we have hooks installed if we're testing a django app.
+    # pylint: disable=unused-import
+    import appmap.django  # noqa: F401
+except ImportError:
+    pass  # probably no django
+
 
 class FuncItem:
     def __init__(self, pyfuncitem):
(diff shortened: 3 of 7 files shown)


appmap/_implementation/generation.py
```diff
@@ -42,19 +42,22 @@ def __init__(self, e):
...
 
-        entry = children.setdefault(class_, ClassEntry(class_))
-        children = entry.children
-
-        loc = f'{e.path}:{e.lineno}'
-        children.setdefault(loc, FuncEntry(e))
+            loc = f'{e.path}:{e.lineno}'
+            children.setdefault(loc, FuncEntry(e))
+        except AttributeError:
+            continue
```
apotterri, line 60: This could use a comment to indicate when `AttributeError` gets raised.

appmap/django.py
```diff
@@ -0,0 +1,46 @@
...
+        try:
+            return execute(sql, params, many, context)
+        finally:
+            if self.recorder.enabled:
+                stop = time.monotonic()
+                duration = stop - start
+                # Note this logic is based on django.db.backends.utils.CursorDebugWrapper.
+                if many:
+                    # Sometimes the same query is executed with different parameter sets.
+                    # Instead of substituting them all, just say how many times it was run.
```
apotterri, line 25: Can you explain this a little more? I'm not sure there's a change needed for the comment, I'd just like to know what's going on.

appmap/django.py
```diff
@@ -0,0 +1,46 @@
...
+                    try:
+                        times = len(params)
+                    except TypeError:
+                        times = '?'
+                    sql = '%s times %s' % (times, sql)
```
apotterri, line 30:
```suggestion
                    sql = f'{times} times {sql}'
```

appmap/pytest.py
```diff
@@ -120,8 +127,8 @@ def pytest_pyfunc_call(pyfuncitem):
     path = os.path.join(session.appmap_path, fname)
 
     def write_recording(r):
-        with open(path, 'wb') as appmap:
-            appmap.write(generation.dump(r, metadata))
+        with open(path, 'wb') as appmap_file:
```
apotterri, line 130: AppMaps are only ever files, so this looks like a stutter.

appmap/test/test_django.py
```diff
@@ -0,0 +1,23 @@
...
+    recorder.events().clear()
+    recorder.enabled = True
+
+    django.conf.settings.configure(DATABASES={'default': {'ENGINE': 'django.db.backends.sqlite3', 'NAME': ':memory:'}})
+
+    conn = django.db.connections['default']
+    conn.cursor().execute('SELECT 1').fetchall()
+
+    recorder.enabled = False
+    assert recorder.events()[0].sql_query['sql'] == 'SELECT 1'
```
apotterri, line 23:
```suggestion
    with recorder:
        django.conf.settings.configure(DATABASES={'default': {'ENGINE': 'django.db.backends.sqlite3', 'NAME': ':memory:'}})

        conn = django.db.connections['default']
        conn.cursor().execute('SELECT 1').fetchall()
    assert recorder.events()[0].sql_query['sql'] == 'SELECT 1'
```

.travis.yml
```diff
@@ -7,18 +7,18 @@ stages:
 - deploy
 cache:
   pip: true
-  
+
 install:
 - pip install --upgrade pip
 - pip install poetry
-- poetry install
+- poetry install -E django
```
apotterri, line 14: I wonder if we really want to make `django` optional? What's the downside to always installing and testing it?

apotterri (changes requested):
Thanks for this, it's really exciting to see SQL getting captured.

A couple of questions and minor changes, then should be good to go.

apotterri:
You mentioned in Discord that you needed to update function wrapping to fix a problem when mapping `django`: https://discord.com/channels/766016904056930325/776157929354690580/810895675071529020 .

Does this PR need that change?

appmap/pytest.py
dividedmind: This is to make pylint shut up about naming conflict with `appmap` module.

appmap/django.py
dividedmind: Sometimes the same SQL will get executed with different parameters, so the arguments will be like eg. `sql="INSERT INTO foo VALUES (%s, %s)", params=[["first", "record"], ["second", "record"]]`. I'd assume that the length of `params` could be significant in this case; `params` could be an iterable without known length, hence the `TypeError` exception.

apotterri: Ah, ok, I understand.

I wonder if there's value in rendering the first few`params`, truncating the string if it's too long?

appmap/test/test_django.py
dividedmind: Will this clear the events list? I'm having trouble following the code there.

appmap/django.py
dividedmind: I haven't actually seen this used yet; admittedly, I haven't tried a lot of code, but it could be only rarely used. I just went with what CursorDebugWrapper does; it's what django itself uses to log queries.

.travis.yml
dividedmind: None really, except conceptually it didn't seem like you should NEED django to work on this module, and it's always a bit more to install.

dividedmind:
> You mentioned in Discord that you needed to update function wrapping to fix a problem when mapping `django`: https://discord.com/channels/766016904056930325/776157929354690580/810895675071529020 .
> 
> Does this PR need that change?

Nope, I ended up just using it to figure out where I need to hook.

appmap/test/test_django.py
apotterri: Sorry, my mistake, this should actually start with
```
    from appmap import Recording
    ....
    recording = Recording()
    with recording:
```
`Recording` is a context manager, not `Recorder`, and it's intended to be a public interface.

Your question points to another problem, though: since `Recorder` is a singleton, it probably needs to protect `self._events` better. It should disallow multiple calls to `start_recording` with intervening calls to `stop_recording`, and it should clear out `_events` when `stop_recording` is called.

I don't think this test should need to worry about either of those cases, though. Are you seeing extra events in your appmap?

.travis.yml
dividedmind: I just imagined I'm an outside contributor and wanted to fix an unrelated bug. I'd be surprised that I need to install django for that. django alone pulls four packages; now imagine if we had more of these integrations.

apotterri: I think it should be required, and that the tests in `test_django` should always run. Otherwise, you won't find out that you've broken something in the `django` integration until your Travis build runs

appmap/test/test_django.py
dividedmind: The tests were broken in some orderings (see https://travis-ci.com/github/applandinc/appmap-python/builds/217319944 for instance). I'll see if I can fix that.

apotterri: Try putting your test in a class that inherits from `AppMapTestBase`. It has a `setup_method` member that properly reinitializes the modules.

dividedmind: Ok, so another problem is that using `Recording()` needs an environment variable set. I can see you do that in the other tests, however this strikes me as rather brittle. I have to wonder: surely if you're doing `with Recording` as an external client (seeing as it's a public interface), you really do want to record, and be surprised if you need to set an environment variable for it to actually work?

.travis.yml
dividedmind: Ok, I'll keep the importorskip though in case we want to make it optional after all.

appmap/test/test_django.py
apotterri: Yeah, that's probably true. But, let's leave it for now, and improve it going forward.

dividedmind: And, using `Recording` also requires setting up a config file etc. It becomes more of an integration test that way. `Recording` is already tested with a separate test. Would it be okay to just leave this as it is?

apotterri: Yes, this is fine for now.

apotterri (approved):

PR #39, opened 2021-02-19
HTTP request and response capture in Django

### LINKED ISSUE #27: Add django integration to generate HTTP server events
Both [http_server_request](https://github.com/applandinc/appmap#http-server-request-attributes) and [http_server_response](https://github.com/applandinc/appmap#http-server-response-attributes) should get generated.

dividedmind:
The middleware injection is a bit inelegant, but it works pretty well.

#5 wasn't actually blocking this; unittest cases driven by pytest still activate some hooks, just not the one we were using. I've changed the hook to work with them.

Fixes #27.

### DIFF
--- a/appmap/pytest.py
+++ b/appmap/pytest.py
@@ -113,13 +113,13 @@ def pytest_runtestloop(session):
 
 
 @pytest.hookimpl(hookwrapper=True)
-def pytest_pyfunc_call(pyfuncitem):
+def pytest_runtest_call(item):
     if not env.enabled():
         yield
         return
 
-    session = pyfuncitem.session
-    item = FuncItem(pyfuncitem)
+    session = item.session
+    item = FuncItem(item)
     metadata = dict(session.appmap_metadata)
     metadata.update(item.metadata)
     logger.debug('pytest_pyfunc_call, metadata %s', repr(metadata))
--- a/appmap/test/test_django.py
+++ b/appmap/test/test_django.py
@@ -1,11 +1,14 @@
+# flake8: noqa: E402
+# pylint: disable=unused-import, redefined-outer-name
+
 import pytest
 from appmap._implementation.recording import Recorder
 
 pytest.importorskip('django')
 
-# flake8: noqa: E402
 import django.conf
 import django.db
+import django.test
 import appmap.django  # noqa: F401
 
 django.conf.settings.configure(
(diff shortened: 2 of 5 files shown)


appmap/pytest.py
```diff
@@ -113,13 +113,13 @@ def pytest_runtestloop(session):
 
 
 @pytest.hookimpl(hookwrapper=True)
-def pytest_pyfunc_call(pyfuncitem):
+def pytest_runtest_call(item):
```
apotterri, line 116: Was there some reason this needed to be changed? Hooking`pytest_pyfunc_call` seemed like it would best guarantee that we avoided instrumenting any setup or teardown code.

appmap/test/test_django.py
```diff
@@ -1,23 +1,55 @@
...
-    import django.conf
-    import django.db
-    import appmap.django  # noqa: F401
 
-    recorder = Recorder()
-    recorder.events().clear()
-    recorder.enabled = True
+@pytest.fixture
+def events(recorder):
+    return recorder.events
```
apotterri, line 32: Maybe this should be `yield events` instead?

appmap/pytest.py
dividedmind: The problem is `pytest_pyfunc_call` doesn't get called for unittest cases. (BTW, unfortunately `pytest_runtest_call` does end up encompassing setup and teardown for unittest cases. I suppose for now it's good enough. We'll improve it in #5.)

appmap/test/test_django.py
dividedmind: Why? There's no cleanup that needs to be done.

appmap/pytest.py
apotterri: Is there something else in this PR that depends on this change? Or is this just a convenience? If the latter, is there a way we could make it opt-in instead?

appmap/test/test_django.py
apotterri: I think it's confusing the way it is, and would be more idiomatic if it used `yield`. The fact that `events` isn't a property, when it almost certainly should be, contributes to the confusion.

So, if you added `@property` to `Recorder.events`, you could say

```
@pytest.fixture
def events(recorder):
    yield recorder.events
```
and users of the fixture could just reference `events` with having to call it as a function.

appmap/pytest.py
dividedmind: Without this change HTTP request capture would be pretty much dead code, since *django.test.TestCase* (and I assume what most people use to test HTTP requests) is a *unittest* subclass.

appmap/test/test_django.py
dividedmind: I agree that maybe it should be a property, however I fail to see how using yield would make any difference or how it would be more idiomatic. I suspect one of us (probably me) is misunderstanding how python coroutines (or whatever it's called that uses yield) and pytest fixtures work. I just followed pytest fixture documentation, which uses return for fixtures normally unless they need cleanup after usage.

dividedmind: Anyway, I've reworked the fixture a little bit, hopefully you'll like it better ;)

apotterri: I do like it better, thanks. :)

I guess my issue wasn't really `yield` versus `return`, it was that the output from the fixture was a function. Now that you've eliminated that, it looks good to me.

apotterri (approved):
Thanks for this, another big step forward for us.

PR #41, opened 2021-02-20
Relax python version requirement to 3.5

apotterri:
Eliminate all uses of features added after Python version 3.5. Add `vermin` to help weed out issues, and `tox` to test against supported versions.

There are also a couple of bug fixes here. I don't love the shenanigans in 6f2a9f1. The problem has proved resistant to my efforts to debug it, though, and the current workaround seems good enough for now. It should probably be revisited sometime soon, though.

### DIFF
--- a/appmap/pytest.py
+++ b/appmap/pytest.py
@@ -128,7 +128,7 @@ def pytest_runtest_call(item):
     path = os.path.join(session.appmap_path, fname)
 
     def write_recording(r):
-        with open(path, 'wb') as appmap_file:
+        with open(path, 'w') as appmap_file:
             appmap_file.write(generation.dump(r, metadata))
         logger.info('wrote recording %s', path)
 
--- a/appmap/pytest.py
+++ b/appmap/pytest.py
@@ -1,4 +1,3 @@
-import importlib.metadata
 import logging
 import pytest
 import os
@@ -42,9 +41,10 @@ class FuncItem:
 
     @property
     def feature_group(self):
-        return inflection.humanize(
-            inflection.titleize(self.class_name.removeprefix('Test'))
-        )
+        test_name = self.class_name
+        if test_name.startswith('Test'):
+            test_name = test_name[4:]
+        return inflection.humanize(inflection.titleize(test_name))
 
     @property
     def feature(self):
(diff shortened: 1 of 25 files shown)


appmap/pytest.py
```diff
@@ -35,9 +34,10 @@ def is_in_class(self):
 
     @property
     def feature_group(self):
-        return inflection.humanize(
-            inflection.titleize(self.class_name.removeprefix('Test'))
-        )
+        test_name = self.class_name
```
virajkanwade, line 37: @apotterri since we already have written code for Py3.9, I think rather than back porting the whole code, we should add backward compatibility with wrapper functions. Somewhat similar to six https://pypi.org/project/six/.

Each compatibility function could have notes about the minimum version. As and when we keep moving ahead with version of Python, it would be very easy to keep upgrading code. And it would reduce the technical debt.

virajkanwade: https://docs.python.org/3/whatsnew/3.9.html

> Python 3.9 is the last version providing those Python 2 backward compatibility layers, to give more time to Python projects maintainers to organize the removal of the Python 2 support and add support for Python 3.9.

> Aliases to Abstract Base Classes in the collections module, like collections.Mapping alias to collections.abc.Mapping, are kept for one last release for backward compatibility. They will be removed from Python 3.10.

apotterri: That seems like a reasonable approach. We're under some time pressure to support older versions of Python, though. At this point, I've made all the changes to move use back to 3.5, so I'm inclined to stick with these changes.

appmap/test/conftest.py
```diff
@@ -0,0 +1,14 @@
...
+@pytest.fixture
+def tmp_git(tmp_path):
+    g = utils.git(cwd=tmp_path)
+    g('init')
+    utils.subprocess_run(['touch', 'README.md'], cwd=tmp_path)
+    g('add README.md')
+    g('commit -m "initial commit"')
+    g('remote add origin https://www.example.com/repo.git')
+    utils.subprocess_run(['touch', 'new_file'], cwd=tmp_path)
+    yield g
```
dividedmind, line 14: You can just return g here :)

apotterri: lol, will do.

dividedmind (approved):
Looks good, thanks! I'd just add one comment so that people don't have to go hunting git blame to understand the `cache_clear` thing.

PR #45, opened 2021-02-24
Unittest integration

### LINKED ISSUE #5: Add unittest integration

dividedmind:
Fixes #5.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,16 +9,16 @@ of code and key data flows.
 
 There are several ways to record AppMaps of your Python program using the `appmap` package:
 
-* Run your tests (pytest, unittest[coming soon]) with the environment variable
+* Run your tests (pytest, unittest) with the environment variable
   `APPMAP=true`. An AppMap will be generated for each test case.
-    
+
 * Use the `appmap.record` [context manager](#context-manager) to control recording. The context manager takes
   an instance of an `appmap.Recording`, which can be used to generate the AppMap.
 
 * [coming soon] Run your application server with AppMap remote recording enabled, and use
   the [AppLand browser extension](https://github.com/applandinc/appland-browser-extension)
   to start, stop, and upload recordings.
-  
+
 Once you have made a recording, there are two ways to view automatically generated
 diagrams of the AppMaps.
 
@@ -26,7 +26,7 @@ The first option is to load the diagrams directly in your IDE, using the [AppMap
 for VSCode](https://marketplace.visualstudio.com/items?itemName=appland.appmap).
 
 The second option is to upload them to the [AppLand server](https://app.land) using the
-[AppLand CLI](https://github.com/applandinc/appland-cli/releases).  
+[AppLand CLI](https://github.com/applandinc/appland-cli/releases).
 
 ### Supported versions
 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -69,15 +69,15 @@ configuration excludes `app.mod2.MyClass`
...
   `WARNING`, `INFO`, `DEBUG`. Not case-sensitive, defaults to `WARNING`.
-  
+
 * `APPMAP_OUTPUT_DIR` specifies the root directory for writing AppMaps. Defaults to
   `tmp/appmap`.
-  
+
 ## Test Frameworks
-`appmap-python` supports recording [pytest](https://pytest.org) and `unittest`[coming soon]
+`appmap-python` supports recording [pytest](https://pytest.org) and `unittest`
```
apotterri, line 80: Even though it's in the standard library, it's probably worth putting a link to the unittest doc here. I'd say linking to the 3.9 page is fine.

README.md
```diff
@@ -118,9 +118,10 @@ PASSED
 ====================================================================== 1 passed in 0.45s ======================================================================
 ```
 
-### unittest [coming soon]
+### unittest
 `import appmap.unittest`. Instruments subclasses of `unittest.TestCase` and records each
-`test_*` function in the subclasses.
+`test_*` function in the subclasses. You can also use `python -m appmap.unittest` exactly like
```
apotterri, line 123: Nice!

appmap/_implementation/testing_framework.py
```diff
@@ -0,0 +1,123 @@
...
+class session:
+    def __init__(self, name, version=None):
+        if not env.enabled():
+            return
+
+        self.appmap_path = Path(configuration.Config().output_dir) / name
+        self.appmap_path.mkdir(parents=True, exist_ok=True)
+
+        # the tests might chdir so make sure to keep the path absolute
+        self.appmap_path = self.appmap_path.resolve()
```
apotterri, line 95: As of #44, Config.output_dir is guaranteed to be an absolute path, so you don't need to do this.

(If somehow you're seeing different behavior, that's a bug, and we should fix it.)

appmap/_implementation/testing_framework.py
```diff
@@ -0,0 +1,123 @@
+import inflection
```
apotterri, line 1: Did you have to make any changes to this code after you copied it out of pytest.py?

apotterri (changes requested):
This is looking good, just a couple of minor changes. 

The Travis build is red, though, so that needs to get sorted out. (FWIW, I confirmed that the tests fail locally for 3.9.1, too.)

Thanks!

appmap/_implementation/testing_framework.py
dividedmind: Ah, ok, thanks.

appmap/_implementation/testing_framework.py
dividedmind: `FuncItem` was mostly unmodified, except I changed it so it takes generic `cls`, `name` and `method_id` arguments instead of the pytest-specific item. I've simplified the session setup code a bit (using `pathlib` instead of `os`, and generalizing it) and especially the recording part -- I changed it to be a context manager, and instead of using `exit_hook` I simply write the appmap after `yield` continuation. I've found the hook approach a bit confusing and unnecessary.

dividedmind:
This should fix the build, fingers crossed.

dividedmind:
Man, that 3.5 is a drag. Are we sure [5 months ago](https://endoflife.date/python) isn't long enough to stop supporting it? ;)

apotterri (approved):

PR #49, opened 2021-02-26
removed full base path and only show relative path for events

ryanmarshall:
This will take out full root path and only leave relative path of running app. Would love to discuss if i missed anything or didn't know about a specific consideration

### DIFF
--- a/appmap/_implementation/event.py
+++ b/appmap/_implementation/event.py
@@ -1,5 +1,7 @@
 import inspect
 import threading
+import re
+import os 
 from itertools import chain
 from functools import partial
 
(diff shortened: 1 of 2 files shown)


appmap/_implementation/event.py
```diff
@@ -70,9 +72,12 @@ def make(fn, isstatic):
         introspecting the given function.
         """
         defined_class, method_id = split_function_name(fn)
+        app_root = os.path.dirname(os.path.realpath(__file__))
```
apotterri, line 75: `__file__` is the location of this source file, i.e. the path to `event.py`, which will be a subdirectory under the directory where the `appmap-python` package is installed.

There's no reason for this to be related to the directory where the source file for `fn` is located. Instead, we need to be capturing the absolute path to the current working directory on [startup](https://github.com/applandinc/appmap-python/blob/master/appmap/_implementation/configuration.py#L46), and stripping that from the path.

You'll also need to add at least one test to show that this functionality works.

apotterri (changes requested):

PR #51, opened 2021-02-26
Make sure wrap_exec_module doesn't rewrap

apotterri:
Add an attribute to a wrapped exec_module to make sure
wrap_exec_module doesn't rewrap a metapath finder's exec_module.

### DIFF
--- a/appmap/_implementation/recording.py
+++ b/appmap/_implementation/recording.py
@@ -218,6 +218,11 @@ class Recorder:
 
 
 def wrap_exec_module(exec_module):
+    marker = '_appmap_exec_module_'
+    if getattr(exec_module, marker, None):
+        logger.debug('%s already wrapped', exec_module)
+        return exec_module
+
     @wraps(exec_module)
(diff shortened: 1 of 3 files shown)


appmap/_implementation/recording.py
```diff
@@ -229,7 +234,8 @@ def wrapped_exec_module(*args, **kwargs):
                      args,
                      kwargs)
         exec_module(*args, **kwargs)
-        Recorder().do_import(*args, **kwargs)
+        import_fn(*args, **kwargs)
```
dividedmind, line 237: Is passing this as argument solely for testing? If so, could we monkey patch instead?

apotterri: Yes, it's solely for testing. In my haste to get this finished, I missed `monkeypatch.setattr`, and the alternatives were all more complicated than I wanted.

I changed it to use `monkeypatch.setattr`.

apotterri:
@dividedmind this is ready for another look. Thanks!

dividedmind (approved):

PR #52, opened 2021-03-01
Write metadata even if git is not found

### LINKED ISSUE #54: Failure writing AppMap when git is unavailable
If git is not on the PATH, writing the `metadata` section to the AppMap will fail.

apotterri:
Fixes #54 .


kgilpin (approved):

appmap/_implementation/utils.py
```diff
@@ -61,16 +60,35 @@ class name and the method name.
...
-    Ret = namedtuple('Ret', ['returncode', 'stdout'])
     try:
         out = subprocess.check_output(command_args,
+                                      stderr=subprocess.STDOUT,
                                       cwd=str(cwd), universal_newlines=True)
-        return Ret(stdout=out, returncode=0)
+        return ProcRet(stdout=out, returncode=0)
     except subprocess.CalledProcessError as exc:
-        return Ret(stdout=exc.stdout, returncode=exc.returncode)
+        return ProcRet(stderr=exc.stdout, returncode=exc.returncode)
```
dustinbyrne, line 90: Just to confirm: we're storing `stdout` as `stderr` here

dustinbyrne (approved):
LGTM. I left one minor comment.

appmap/_implementation/utils.py
apotterri: Yeah, when an error has happened. `check_output` doesn't have a great interface -- we'll get rid of it as soon as we can.

PR #57, opened 2021-03-05
Support shallow capture, package specs

### LINKED ISSUE #48: Add support for specifying a Python package in appmap.yml
Analogous to `appmap-ruby`'s `gem` support: https://github.com/applandinc/appmap-ruby#configuration .

```yaml
name: my-app
packages:
- package: pyyaml
```
would add the modules from the `pyyaml` package, and turn on shallow mode for the modules.

dividedmind:
Fixes #48 .

Reviewing commit-by-commit recommended.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,21 +46,38 @@ pip install appmap
 For projects that use `poetry` , add the `appmap` package to `pyproject.toml`.
 
 ## Configuration
-Add your modules as `path` entries in `appmap.yml`:
+Add your modules as `path` entries in `appmap.yml`, and external packages
+(distributions) as `dist`:
 
 ```yaml
 name: my_python_app
 packages:
 - path: app.mod1
+  shallow: true
 - path: app.mod2
   exclude:
   - MyClass
   - MyOtherClass.my_instance_method
   - MyOtherClass.my_class_method
+- dist: Django
+  exclude:
+  - django.db
 ```
 
 Note that `exclude`s are resolved relative to the associated path. So, for example, this
-configuration excludes `app.mod2.MyClass`
+configuration excludes `app.mod2.MyClass`.
+
+For external [distribution packages](https://packaging.python.org/glossary/#term-Distribution-Package)
+use the `dist` specifier; the names are looked up in the
+[database of installed Python distributions](https://www.python.org/dev/peps/pep-0376/).
+This is generally the same package name as you'd give to `pip install` or put
+in `pyproject.toml`. You can additionally use `path` and `exclude` on `dist`
+entries to limit the capture to specific patterns.
+
+Note by default shallow capture is enabled on `dist` packages, supressing tracking
+of most internal execution flow, which allows you to capture the interaction without
+getting bogged down with detail. If this isn't what you want, use `shallow: false`.
+You can also use `shallow: true` on `path` entries.
 
 ## Environment Variables
 
--- a/appmap/_implementation/configuration.py
+++ b/appmap/_implementation/configuration.py
@@ -2,6 +2,8 @@
 Manage Configuration AppMap recorder for Python.
 """
 
+import importlib_metadata
+import inspect
 import logging
 import os.path
 
@@ -72,7 +74,9 @@ def startswith(prefix, sequence):
 class PathMatcher:
     def __init__(self, prefix, excludes=None, shallow=False):
         excludes = excludes or []
-        self.prefix = prefix.split('.')
+        self.prefix = []
+        if prefix:
+            self.prefix = prefix.split('.')
         self.excludes = [x.split('.') for x in excludes]
         self.shallow = shallow
 
@@ -83,11 +87,36 @@ class PathMatcher:
             result = not any(startswith(x, name) for x in self.excludes)
         else:
             result = False
-        logger.debug('PathMatcher.matches(%r, %r) -> %r', self, obj, result)
+        logger.debug('%r.matches(%r) -> %r', self, obj, result)
         return result
 
     def __repr__(self):
         return 'PathMatcher(%r, %r, shallow=%r)' % (
+            '.'.join(self.prefix),
+            ['.'.join(ex) for ex in self.excludes],
+            self.shallow
+        )
+
+
+class DistMatcher(PathMatcher):
+    def __init__(self, dist, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.dist = dist
+        self.files = [str(pp.locate()) for pp in importlib_metadata.files(dist)]
+
+    def matches(self, obj):
+        try:
+            logger.debug('%r.matches(%r): %s in %r', self, obj, inspect.getfile(obj), self.files)
+            if inspect.getfile(obj) not in self.files:
+                return False
+        except TypeError:
+            # builtins don't have file associated
+            return False
+        return super().matches(obj)
+
+    def __repr__(self):
+        return 'DistMatcher(%r, %r, %r, shallow=%r)' % (
+            self.dist,
             '.'.join(self.prefix),
             ['.'.join(ex) for ex in self.excludes],
             self.shallow
@@ -128,14 +157,27 @@ class MatcherFilter(Filter):
         return next((m for m in self.matchers if m.matches(fn)), None)
 
 
+def matcher_of_config(package):
+    dist = package.get('dist', None)
+    if dist:
+        return DistMatcher(
+            dist,
+            package.get('path', None),
+            package.get('exclude', []),
+            shallow=package.get('shallow', True)
+        )
+    return PathMatcher(
+        package['path'],
+        package.get('exclude', []),
+        shallow=package.get('shallow', False)
+    )
+
+
 class ConfigFilter(MatcherFilter):
     def __init__(self, *args, **kwargs):
         matchers = []
         if Env.current.enabled:
-            matchers = [PathMatcher(
-                p['path'], p.get('exclude', []),
-                shallow=p.get('shallow', False)
-            ) for p in Config().packages]
+            matchers = [matcher_of_config(p) for p in Config().packages]
         super().__init__(matchers, *args, **kwargs)
 
 
--- a/appmap/_implementation/configuration.py
+++ b/appmap/_implementation/configuration.py
@@ -36,7 +36,7 @@ class Config:
 
         config_file = os.getenv("APPMAP_CONFIG", "appmap.yml")
         with open(config_file) as file:
-            self._config = yaml.load(file, Loader=yaml.BaseLoader)
+            self._config = yaml.safe_load(file)
             logger.debug('self._config %s', self._config)
 
         self._initialized = True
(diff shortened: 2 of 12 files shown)


apotterri:
The `excludes` behavior, and the dependency it introduces between filters, was intentional. One of the things that's missing from `appmap-ruby`'s configuration is the ability to exclude built-in functions via `appmap.yml`. This seemed like a way to add that functionality here: if the filter can make a decision about the function (either to include or exclude), then processing stops at that point. Only if it can't decide should it forward to the next filter.

I think this functionality would be good to have, so maybe we can bring it back in a way that's less surprising?

dividedmind:
> The `excludes` behavior, and the dependency it introduces between filters, was intentional. One of the things that's missing from `appmap-ruby`'s configuration is the ability to exclude built-in functions via `appmap.yml`. This seemed like a way to add that functionality here: if the filter can make a decision about the function (either to include or exclude), then processing stops at that point. Only if it can't decide should it forward to the next filter.
> 
> I think this functionality would be good to have, so maybe we can bring it back in a way that's less surprising?

I'm not sure I understand exactly what you want to achieve. Could you give me an example of excluding built-in functions like that? (And if it was intentional, why wasn't there a test case for that? ;) )

kgilpin:
Actually, you can exclude builtin functions in appmap-ruby now. There is a
top-level "exclude" key where you can exclude by class and function name

https://github.com/applandinc/appmap-ruby/#configuration

apotterri:
There's no test case for it because there aren't any built-in functions to exclude yet.

By default, `appmap-ruby` instruments quite a number of built-in functions, and I anticipate that `appmap-python` will, too. I think there needs to be a way for the user to say, "I don't want this particular built-in to be instrumented". The behavior I implemented for `excludes` was to support that.

We're going to want to support the `excludes` `appmap.yml` attribute at some point (probably when we add instrumentation of built-in functions), so I guess this can get deferred until then.

kgilpin:
May want to call it "exclude" for consistency with Ruby

appmap/_implementation/configuration.py
```diff
@@ -55,147 +56,137 @@ def packages(self):
...
-    To determine whether a function is in one of the sets, the
-    fully-qualified name of the function is considered, including the
-    packages(s), module, and class that contain it.  As an example,
-    consider a function named `pkg1.pkg2.mod1.Class1.func`.  Any of
-    `pkg1`, `pkg1.pkg2.`, `pkg1.pkg2.mod1`, `pkg1.pkg2.mod1.Class1`,
-    or `pkg1.pkg2.mod1.Class1.func` will match this function (causing
-    it to be included or excluded, as appropriate).  Note that only
-    full names will match.  (This is achieved in the code by appending
-    `.` to the name retrieved from the config file.)
+    Check if needle is a prefix of haystack.
```
apotterri, line 69: What are the types of needle and haystack? Maybe they could use more descriptive names?

appmap/_implementation/configuration.py
```diff
@@ -55,147 +56,137 @@ def packages(self):
...
-    logger.debug(('function_in_set, class_name %s'
-                  ' fnname %s'
-                  ' which %s'
-                  ' ret %s'),
-                 class_name,
-                 fn_name,
-                 which,
-                 ret)
-    return ret
+def splitname(obj):
```
apotterri, line 59: Seems like `splitname` could be useful in other contexts as well. Maybe it belongs in `utils.py`?

appmap/_implementation/configuration.py
```diff
@@ -37,7 +38,7 @@ def __init__(self):
 
         config_file = os.getenv("APPMAP_CONFIG", "appmap.yml")
         with open(config_file) as file:
-            self._config = yaml.load(file, Loader=yaml.BaseLoader)
+            self._config = yaml.safe_load(file)
```
apotterri, line 41: 👍

appmap/_implementation/configuration.py
```diff
@@ -55,147 +56,137 @@ def packages(self):
...
+            ['.'.join(ex) for ex in self.excludes],
+            self.shallow
+        )
+
+
+class DistMatcher(PathMatcher):
+    def __init__(self, dist, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        self.dist = dist
+        self.files = [str(pp.locate()) for pp in importlib_metadata.files(dist)]
```
apotterri, line 105: Why do the matching on file rather than on module?

apotterri (changes requested):
This is looking good. Can't wait to get it into a release!

appmap/_implementation/configuration.py
dividedmind: Because packages don't provide modules, packages provide files. One package can provide many modules and the same module can be provided by many packages (in which case of course only one implementation can be loaded at the same time, at least at the root namespace). A module doesn't know its package; in fact, it doesn't know about packages at all (and neither does python core). There's no way to map from module to package except by matching its file path to the package manifest.

appmap/_implementation/configuration.py
dividedmind: Perhaps, although I thought it makes more sense to keep it near the client code until and unless there's another usage. `utils` packages are always a bit of annoying indirection when reading code (and I've read good arguments that they're bad style, though I'm not entirely convinced).

appmap/_implementation/configuration.py
dividedmind: Any iterable will work; what would you suggest? I thought they're pretty descriptive, and in any case, the function is simple enough that it's quite easy to read to see.

README.md
```diff
@@ -46,21 +46,36 @@ pip install appmap
...
+- dist: Django
+  exclude:
+  - django.db
 ```
 
 Note that `exclude`s are resolved relative to the associated path. So, for example, this
-configuration excludes `app.mod2.MyClass`
+configuration excludes `app.mod2.MyClass`.
+
+`dist` specifiers use importlib.metadata, and so should work regardless of
```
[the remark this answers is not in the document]
dividedmind: Perhaps, but packages and distributions are not the same thing; distributions do not always come from packages and packages not always contain exactly one distribution. Python doesn't know about packages and there are many tools and formats of packaging, and probably more will be invented. I don't think it's a good idea to interact with specific packaging tools; thing that Python knows about is a "distribution" and calling it something else may lead to confusion ;)

README.md
```diff
@@ -46,21 +46,36 @@ pip install appmap
...
+the package flavor (egg, wheel, etc.). You can additionally use `path` and `exclude`
+on `dist` modules to limit the capture to specific patterns.
+
+Note by default shallow capture is enabled on `dist` packages, supressing tracking
+of most internal execution flow, which allows you to capture the interaction without
+getting bogged down with detail. If this isn't what you want, use `shallow: false`.
+You can also use `shallow: true` on `path` packages, but the implementation isn't
+foolproof so YMMV.
```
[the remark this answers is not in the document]
dividedmind: Fair enough. If someone's curious they can always dive into source code :)

README.md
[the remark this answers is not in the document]
apotterri: Isn't the value of `package` what I said: what  you'd pass to `pip install`, or put in `pyproject.toml` as a dependency? 

Maybe my confusion is understandable: https://pypi.org/ , https://packaging.python.org/glossary/#term-Distribution-Package ? ;) Also:
```sh
ajp@Alans-MacBook-Pro appmap-python % pip --help

Usage:
  pip <command> [options]

Commands:
  install                     Install packages.
  download                    Download packages.
  uninstall                   Uninstall packages.
  freeze                      Output installed packages in requirements format.
  list                        List installed packages.
```

So, while `dist` may be technically correct, it seems like most users are probably going to know it this thing as a `package`.

appmap/_implementation/configuration.py
apotterri: The function is called `startswith`. In what sense would a `haystack` start with a `needle`? Maybe just giving an example of usage would clear up my confusion.

appmap/_implementation/configuration.py
apotterri: Yeah, I'm not entirely convinced, either. I think it's ok for some things to wind up in a catch-all. I guess this can get moved the first time somebody wants to write `from .configuration import splitname`....

appmap/_implementation/configuration.py
apotterri: Ok, I understand this now. Thanks.

README.md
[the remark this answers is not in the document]
dividedmind: Do we have a pythonista we could ask? I feel we're both dabbling ;)

appmap/_implementation/configuration.py
dividedmind: :D

appmap/_implementation/configuration.py
dividedmind: > I guess this can get moved the first time somebody wants to write `from .configuration import splitname`....

Exactly my line of thought. YAGNI, but constant refactoring :)

appmap/_implementation/configuration.py
dividedmind: It's a metaphor (perhaps with a bit of a stretch) for finding the needle in a haystack. `prefix` and `collection`? Are we bikeshedding yet?

kgilpin:
Nice PR but after all this discussion I am looking for a Merged message!

apotterri:
We haven't reached an agreement about what the new attribute should be called, `dist` or `package`.

`dist`, while technically correct, isn't the term that's widely used, so I'm concerned that it will lead to confusion.

Mayne it doesn't matter so much? I don't know.

dividedmind:
I'd also note that `package` could look weird given that the top level key is `packages`:
```yaml
packages:
- package: foo
- package: bar
```
looks quite duplicative; it's begging to be changed to `name` and allowed to be shortened to just the string.

README.md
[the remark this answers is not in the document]
apotterri: No, you and I are basically it, at least for quick turnaround.

I guess we can go with `dist`. If we do, though, I think this paragraph needs to be expanded to introduce the concept more gently:
> Also, I think we need to help with user understand what the value ~package~ `dist` can be: it's the string you'd use for `pip instal`l, or you'd add as a dependency in `pyproject.toml`. I know that's what "specifiers use importlib.metadata" implies, but it feels like it would be better to really spell it out.

README.md
```diff
@@ -46,21 +46,36 @@ pip install appmap
...
+  - django.db
 ```
 
 Note that `exclude`s are resolved relative to the associated path. So, for example, this
-configuration excludes `app.mod2.MyClass`
+configuration excludes `app.mod2.MyClass`.
+
+`dist` specifiers use importlib.metadata, and so should work regardless of
+the package flavor (egg, wheel, etc.). You can additionally use `path` and `exclude`
+on `dist` modules to limit the capture to specific patterns.
```
apotterri, line 72: This should be "`dist` packages", no?

dividedmind: I suppose... I think I'll change it to "entries".

apotterri (approved):
Thanks for the updates, looks great. Go ahead and merge whenever you're ready.

PR #59, opened 2021-03-08
Clear events list when starting a recording

dividedmind:
This should fix a problem where unrelated events end up in test appmaps.


apotterri (approved):
LGTM. Thanks for fixing it so quickly.

dividedmind:
> LGTM. Thanks for fixing it so quickly.

NP, I'll merge it once the build passes, but feel free to do it if you notice it first :)

apotterri (approved):

PR #60, opened 2021-03-08
Enable PathMatcher to handle builtin functions

apotterri:


appmap/_implementation/configuration.py
```diff
@@ -61,7 +61,12 @@ def splitname(obj):
     Given an object that has a __module__ and __qualname__,
     return a list of name components from both.
     """
-    return obj.__module__.split('.') + obj.__qualname__.split('.')
+    if getattr(obj, '__module__', None):
+        modname = obj.__module__
+    else:
+        modname = 'builtins'
```
dividedmind, line 67:
```suggestion
    modname = obj.__module__ or 'builtins'
```

appmap/_implementation/configuration.py
```diff
@@ -61,7 +61,12 @@ def splitname(obj):
     Given an object that has a __module__ and __qualname__,
     return a list of name components from both.
     """
```
dividedmind, line 63:
```suggestion
    If __module__ is None, 'builtins' is used.
    """
```

dividedmind (changes requested):
Needs doc change and I have a style suggestion, but thanks for the fix!

PR #63, opened 2021-03-09
Add appmap.labels decorator

### LINKED ISSUE #61: django.dispatch crashes the appmap client.
When recording chipy_org with the Django package included in appmap.yml, the client crashes a test unless django.dispatch is excluded.

apotterri:
Provide the `labels` decorator to allow the user to add annotations
(labels) to the classMap entries for functions.

Also, get rid of some code that modified an instrumented function's kwargs. Fixes #61.

### DIFF
--- a/appmap/_implementation/event.py
+++ b/appmap/_implementation/event.py
@@ -120,7 +120,7 @@ class Param:
 
 class CallEvent(Event):
     __slots__ = ['defined_class', 'method_id', 'path', 'lineno',
-                 'static', 'receiver', 'parameters']
+                 'static', 'receiver', 'parameters', 'labels']
 
     @staticmethod
     def make(fn, fntype):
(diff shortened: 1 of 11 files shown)


dividedmind (approved):
Looks good, thanks! If it's not too much work, please spin off the README formatting changes to a separate commit before merging. And the kwargs fix needs a changelog entry :)

appmap/_implementation/event.py
```diff
@@ -45,6 +45,15 @@ def reset(cls):
...
 
+def describe_value(val):
+    """
+    Return a description of the given value, with the form
+    (class_name, object_id).
+    """
+    return (fqname(type(val)),
+            id(val))
+
+
```
dividedmind, line 56: How about instead returning a `{class, value, object_id}` dict and then use that to DRY up `Param` and `FuncReturnEvent`?

dividedmind (changes requested):
Good catch! I have a suggestion for a slight improvement.

appmap/_implementation/event.py
apotterri: Will do.

dividedmind (approved):
Thanks!

PR #67, opened 2021-03-24
Handle wrapped functions, add test file location to metadata

### LINKED ISSUE #65: Wrapped functions with mismatched signatures cause mapping failures
If the signature of the function returned by the `functools.wraps` decorator doesn't match the signature of the decorated function, mapping fails when the function is called. For example, `unittest.mock.patch` (which uses `functools.wraps`) returns a function with the signature `()` (i.e. no arguments). When called, this function computes the correct arguments before it calls the decorated test function. If the test function has a signature other than `()`, an exception is thrown when we try to retrieve parameters based on the test function's signature.

### LINKED ISSUE #66: Location of test methods is recorded in metadata
When `unittest` and `pytest` tests are recorded, the location of the test method should appear in `metadata.recording.source_location`.

apotterri:
Fixes #65. Fixes #66.


appmap/test/data/example_class.py
```diff
@@ -18,6 +19,18 @@ def method_not_called_directly(self):
         return 'Super#instance_method'
 
 
+def wrap_fn(fn):
+    @wraps(fn)
+    def wrapped_fn(*args, **kwargs):
+        try:
+            print('calling %s' % (fn.__name__))
+            return fn(*args, **kwargs)
+        finally:
```
dividedmind, line 28: That's a nice use of finally clause!

appmap/test/data/unittest/simple/simple.py
```diff
@@ -0,0 +1,9 @@
+class Simple:
+    def hello(self):
+        return 'Hello'
+
+    def world(self):
+        return 'world'
+
+    def hello_world(self, bang):
+        return '%s %s%s' % (self.hello(), self.world(), bang)
```
dividedmind, line 9: It's not very important, but I'll note this could go into `__init__` module, leading to better test names, `simple.Simple` vs. `simple.simple.Simple`. Simpler! ;)

dividedmind (approved):
If it's not too much bother, please make sure code is correctly assigned to commits -- it appears some fixups for the first commit ended up in the second one; alternatively just squash the two.

You can add changelog entries in the respective commits as long as you're updating history anyway. It's always nice when you can `git blame CHANGELOG.md` and see exactly when and how something was added.

Looks good otherwise, thanks!

PR #73, opened 2021-03-30
Reset shallow tracking rule when a function returns

### LINKED ISSUE #72: Only the first of multiple calls to a shallow-tracked packages are recorded
If this code

```python
import yaml
import io

class Example
    def call_dump(self):
        yaml.Dumper(io.StringIO()).open()
        yaml.Dumper(io.StringIO()).open()
```

is recorded like this:

```python
import yaml

import appmap

appmap.instrument_module(yaml)
r = appmap.Recording()
with r:
    from example import Example
    Example().call_dump()
print(appmap.generation.dump(r))
```

using this `appmap.yml`

```yaml
name: example
packages:
- path: example
- dist: PyYAML
```

the resulting recording looks like this:
image

It should show two invocations of `Serializer#open`.

### LINKED ISSUE #74: pytest integration is broken
One of the changes made for flask integration broke the pytest integration. Trying to map flaskbb shows this error:

```
  File "/Users/ajp/.pyenv/versions/3.9.1/envs/flaskbb/lib/python3.9/site-packages/_pytest/assertion/rewrite.py", line 100, in find_spec
    spec = self._find_spec(name, path)  # type: ignore
  File "/Users/ajp/.pyenv/versions/3.9.1/envs/flaskbb/lib/python3.9/site-packages/appmap/_implementation/recording.py", line 259, in wrapped_find_spec
    spec = find_spec(*args, **kwargs)
  File "<frozen importlib._bootstrap_external>", line 1349, in find_spec
  File "<frozen importlib._bootstrap_external>", line 1321, in _get_spec
  File "<frozen importlib._bootstrap_external>", line 1444, in find_spec
AttributeError: 'AssertionRewritingHook' object has no attribute 'rpartition'
```

apotterri:
Improve handling of multiple calls into a shallow-tracked package. Fixes #72. (I split the changes into two separate commits to make it easier to understand how the expected appmap changed, so it's probably easier to review them individually.)

These changes also switch to using `wrapt` for wrapping the metapath finder functions, which fixes #74. It's currently hard to test the exact environment that reproduces #74, so I'll look at adding a test in a follow-up PR.

Here are some more details for changes to shallow tracking:
Given these functions in a class
```python
    @staticmethod
    def call_yaml():
        return ExampleClass.dump_yaml('ExampleClass.call_yaml')

    @staticmethod
    def dump_yaml(data):
        import io
        from yaml import Dumper
        stream = io.StringIO()
        dumper = Dumper(stream)
        try:
            dumper.open()
            dumper.represent(data)
            dumper.close()
        finally:
            dumper.dispose()
        return stream.getvalue()
```
a full recording for the `call_yaml` function looks like this:
image

Without this change, a recording that captures the PyYAML dist in shallow mode looks like this:
image

i.e. only the call to `open` is shown.

With this change, all the calls from `call_yaml` into the PyYAML dist are shown:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Fixed
 - [#72] Multiple calls into a shallow-tracked package are now recorded.
+- [#74] pytest integration works again
 
 ## [0.8.0.dev2] - 2021-03-26
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- [#72] Multiple calls into a shallow-tracked package are now recorded.
+
 ## [0.8.0.dev2] - 2021-03-26
 ### Added
 - [#68] Support `APPMAP_DISPLAY_PARAMS`.
--- a/appmap/_implementation/instrument.py
+++ b/appmap/_implementation/instrument.py
@@ -1,3 +1,4 @@
+from collections import namedtuple
 from contextlib import contextmanager
 from functools import wraps, WRAPPER_ASSIGNMENTS
 import logging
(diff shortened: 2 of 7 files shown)


apotterri:
@dividedmind Given the severity of the pytest bug, Kevin suggested that I get it merged and up on PyPI. I'll be happy to address any concerns you have in a follow-up PR.

PR #78, opened 2021-04-13
feat: Record HTTP server request parameters as `message`

dustinbyrne:
image

### DIFF
--- a/appmap/test/test_django.py
+++ b/appmap/test/test_django.py
@@ -48,3 +48,85 @@ def test_http_capture(events):
         'status_code': 404,
         'mime_type': 'text/html'
     }
+
+
+def test_message_capture_get(events):
+    client = django.test.Client()
+    client.get('/test', { 'my_param': 'example' }, content_type='application/x-www-form-urlencoded')
+
+    assert events[0].message == [
+        {
+            'name': 'my_param',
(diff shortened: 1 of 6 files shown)


apotterri (changes requested):
This looks good, except the tests fail for python 3.5. Once that gets sorted out, it'll be good to go.

apotterri (dismissed):

apotterri (changes requested):
Oops, there is one more thing needed here. Can you add a CHANGELOG entry, please?

appmap/test/test_django.py
```diff
@@ -48,3 +48,85 @@ def test_http_capture(events):
         'status_code': 404,
         'mime_type': 'text/html'
     }
+
+
+def test_message_capture_get(events):
+    client = django.test.Client()
+    client.get('/test', { 'my_param': 'example' }, content_type='application/x-www-form-urlencoded')
```
dividedmind, line 55: `GET` requests shouldn't take `content_type` as it relates to body, which GET requests lack.

dividedmind (approved):
Just one small note, plus missing changelog entry.

apotterri:
Merged here: 2a5be737d7ba8e47782f946c6cb61dd34f147a53

PR #82, opened 2021-04-17
Generate appmaps for unittest tests run by pytest

### LINKED ISSUE #81: unittest tests run by pytest don't get recorded
When pytest is used to run unittest tests (e.g in https://github.com/scrapy/scrapy), no appmaps are generated.

### LINKED ISSUE #94: Drop support for Python 3.5
It's over five years old and seven months after its end of life: https://endoflife.date/python

It breaks builds often and requires frequent special cases and special fixes.

apotterri:
Fixes #81. Fixes #94.

[Comment below is the original PR comment. It's not applicable any more, before we're no longer using `pytest_runtest_call`. @dividedmind was correct, it's not the right hook to use. I'm leaving the comment here for context {ajp 20210422}]

Have the pytest integration hook `pytest_runtest_call` instead of `pytest_pyfunc_call`. The former is called for both `pytest` and `unittest` tests, the latter only for `pytest` tests.

You can see the behavior of the hooks by examining pytest's debug log (when pytest is run with `--debug`). For example, when running the `unittest` test case `test_undefined_parse_method` in [scrapy](https://github.com/scrapy/scrapy), the log shows:

```
    pytest_runtestloop [hook]
        session: <Session scrapy exitstatus=<ExitCode.OK: 0> testsfailed=0 testscollected=1>
      pytest_runtest_protocol [hook]
          item: <TestCaseFunction test_undefined_parse_method>
          nextitem: None
        ...
        pytest_runtest_setup [hook]
            item: <TestCaseFunction test_undefined_parse_method>
        ...
        finish pytest_runtest_setup --> [] [hook]
        ...
        pytest_runtest_call [hook]
            item: <TestCaseFunction test_undefined_parse_method>
        finish pytest_runtest_call --> [] [hook]
```

I've cut out some setup, but this shows that by starting recording in `pytest_runtest_call`, only the test method, and not any of the setup or teardown, will be captured.

Similarly, for our test case `test_recording_works`, run with `pytest`:

```
    pytest_runtestloop [hook]
        session: <Session appmap-python exitstatus=<ExitCode.OK: 0> testsfailed=0 testscollected=1>
      pytest_runtest_protocol [hook]
          item: <Function test_recording_works>
          nextitem: None
        ...
        pytest_runtest_setup [hook]
            item: <Function test_recording_works>
        ...
        finish pytest_runtest_setup --> [] [hook]
        ...
        pytest_runtest_call [hook]
            item: <Function test_recording_works>
          pytest_pyfunc_call [hook]
              pyfuncitem: <Function test_recording_works>
          early skip of rewriting module: example_class [assertion]
          finish pytest_pyfunc_call --> True [hook]
        finish pytest_runtest_call --> [] [hook]

### CODE REVIEW

**Comment by dividedmind** [comment]
> I've cut out some setup, but this shows that by starting recording in pytest_runtest_call, only the test method, and not any of the setup or teardown, will be captured.

I don't see that from the log. What's more, I remember from my work on this that in fact with pytest_runtest_call the setup and teardown do in fact get recorded in unittest cases; AFAIR it even says so in documentation and is evident from pytest source code that it sees unittest cases as a single unit, including setup and teardown.

Why aren't unittest cases recorded by the unittest hook as they used to be?

**Comment by apotterri** [comment]
These snippets from the log make it look to me like the setup hook finishes before pytest_runtest_call is invoked.
When running a `unittest` test case:
```
       pytest_runtest_setup [hook]
            item: <TestCaseFunction test_undefined_parse_method>
        ...
        finish pytest_runtest_setup --> [] [hook]
        ...
        pytest_runtest_call [hook]
```

Same for a pytest test case: 
```
        pytest_runtest_setup [hook]
            item: <Function test_recording_works>
        ...
        finish pytest_runtest_setup --> [] [hook]
        ...
        pytest_runtest_call [hook]
            item: <Function test_recording_works>
```

What are you seeing that makes you think something else is happening?

By "the unittest hook", do you mean: https://github.com/applandinc/appmap-python/blob/master/appmap/unittest.py#L15, or something else? `appmap.unittest` isn't going to be used when you uae `pytest` as the test runner, is it?

**Comment by dividedmind** [comment]
> These snippets from the log make it look to me like the setup hook finishes before pytest_runtest_call is invoked.
> When running a `unittest` test case:
> 
> ```
>        pytest_runtest_setup [hook]
>             item: <TestCaseFunction test_undefined_parse_method>
>         ...
>         finish pytest_runtest_setup --> [] [hook]
>         ...
>         pytest_runtest_call [hook]
> ```

Yes, but that's pytest setup. Unittest setup is called in the `call` part; see note at the bottom of https://docs.pytest.org/en/stable/unittest.html confirming this, or read the source. This is exactly why I reverted to pyfunc hook, as indicated in the comment.

> By "the unittest hook", do you mean: https://github.com/applandinc/appmap-python/blob/master/appmap/unittest.py#L15, or something else? `appmap.unittest` isn't going to be used when you uae `pytest` as the test runner, is it?

Yes. It should be used, that's why https://github.com/applandinc/appmap-python/blob/2a5be737d7ba8e47782f946c6cb61dd34f147a53/appmap/pytest.py#L3 is there. There's even a testcase for this: https://github.com/applandinc/appmap-python/blob/2a5be737d7ba8e47782f946c6cb61dd34f147a53/appmap/test/test_unittest.py#L23

With the proposed change the tests will get recorded twice.

**Comment by apotterri** [comment]
Hmmm. The tests all pass with this change, so maybe `test_unittest_cases` needs to be updated....

I didn't make this change for no reason -- running the tests for scrapy without this change doesn't produce any appmaps, but with the change, they do get generated.

I'll dig in some more and see if I can figure out why. Seems like it would also be good to have a test or two that confirms that setup and teardown don't get recorded.

**Comment by dividedmind** [comment]
> Hmmm. The tests all pass with this change, so maybe `test_unittest_cases` needs to be updated....

Perhaps; it doesn't check the appmap contents currently, just the presence. I suspect with this change the pytest hook overwrites the appmaps generated by the unittest hook. Depending on the specifics of the code, they may or may not be correct and may or may not include setup and/or teardown. (I don't think the test case includes any setup or teardown code.)

> I didn't make this change for no reason -- running the tests for scrapy without this change doesn't produce any appmaps, but with the change, they do get generated.
>
> I'll dig in some more and see if I can figure out why. Seems like it would also be good to have a test or two that confirms that setup and teardown don't get recorded.

Yeah. I guess there must be another another with scrapy specifically preventing the unittest hook from working correctly. However, this change may subtly break appmaps in other projects where it does work correctly and is driven by pytest.

File: appmap/unittest.py [hunk 1]
```diff
@@ -12,23 +13,29 @@ def get_test_location(cls, method_name):
...
 
-def run(self, result=None):
-    method_name = self.id().split('.')[-1]
-    # Use the result's location if provided (e.g. by pytest),
-    # otherwise cobble one together ourselves.
-    if hasattr(result, 'location'):
-        location = result.location
-    else:
-        location = get_test_location(self.__class__, method_name)
+# unittest.case._Outcom.testPartExecutor is used by all supported
```
dividedmind, line 16: Typo ;)

appmap/unittest.py
```diff
@@ -12,23 +13,29 @@ def get_test_location(cls, method_name):
...
+# versions of unittest to run test cases. `isTest` will be True when
+# the part is the actual test method, False when it's setUp or
+# teardown.
+@wrapt.patch_function_wrapper('unittest.case', '_Outcome.testPartExecutor')
+@contextmanager
+def testPartExecutor(wrapped, _, args, kwargs):
+    def _args(test_case, *_, isTest=False, **__):
+        return (test_case, isTest)
+    test_case, is_test = _args(*args, **kwargs)
+    if not is_test:
```
dividedmind, line 26: This gave me pause. Any reason to do it this way instead of using `kwargs.get('isTest')` in the `if` directly, and `test_case = args[0]`? Just wondering, if you find it clearer like this then that's a reason enough :)

appmap/unittest.py
```diff
@@ -12,23 +13,29 @@ def get_test_location(cls, method_name):
...
+        with wrapped(*args, **kwargs):
```
dividedmind, line 27: Is there a way to check or assert that the test isn't already driven by pytest? In case the user ends up somehow importing appmap.unittest.

appmap/pytest.py
```diff
@@ -1,20 +1,59 @@
...
+        # case, it will be a `wrapt` class, rather than just a
+        # function. This is fine: the decorator we apply here will be
+        # called first, starting the recording. Next, the
+        # instrumentation decorator will be called, recording the
+        # `call` event. Finally, the original function will be called,
+        # running the test case. (This nesting of function calls is
+        # verified by the expected appmap in the test for a unittest
+        # TestCase run by pytest.)
+        if hasattr(item, '_testcase'):
+            item.obj = recorded_testcase(item)(item.obj)
```
dividedmind, line 48: The comment mentions `_obj`, however, `obj` is wrapped here. Is this a mistake in the comment or in the code? If the latter, are we missing a test here?

dividedmind (changes requested):
This looks much better, thanks! I just have a couple of suggestions for you to consider.

appmap/unittest.py
apotterri: Fixed.

appmap/unittest.py
apotterri: This way of extracting the arguments is described here: https://wrapt.readthedocs.io/en/latest/decorators.html#processing-function-arguments .

I believe this is necessary to handle the case where the caller of the wrapped function passes the first argument as a kwarg.

appmap/unittest.py
apotterri: I agree that this would be good to do. Let me see if I can find something.

appmap/pytest.py
apotterri: I'm not sure. I'll look at it again and make sure everything lines up.

appmap/unittest.py
dividedmind: Ah, ok, that makes sense. Thanks!

appmap/pytest.py
```diff
@@ -1,20 +1,59 @@
...
 
 if appmap.enabled():
     @pytest.hookimpl
     def pytest_sessionstart(session):
-        """Set up AppMap testing session and associate it with pytest and unittest sessions."""
-        # Use the same session for unittest in case there are any unittest.TestCases.
-        appmap.unittest.session = session.appmap = testing_framework.session(name='pytest', version=pytest.__version__)
+        session.appmap = testing_framework.session(name='pytest', version=pytest.__version__)
+
+    @pytest.hookimpl(hookwrapper=True)
```
dividedmind, line 27: This doesn't need to be a hookwrapper; it doesn't do any cleanup. Just remove this decorator and the yield.

dividedmind:
FWIW, I've just tried this branch on misago and it doesn't record the (unittest, driven by pytest) tests at all.

apotterri:
> FWIW, I've just tried this branch on misago and it doesn't record the (unittest, driven by pytest) tests at all.

Ok, thanks. I'll figure out why before I move forward with these changes.

appmap/pytest.py
apotterri: It was a mistake in the comment. I've fixed it.

apotterri:
> > FWIW, I've just tried this branch on misago and it doesn't record the (unittest, driven by pytest) tests at all.
> 
> Ok, thanks. I'll figure out why before I move forward with these changes.

I ran the misago tests using these changes and saw that appmaps are now being generated.

appmap/unittest.py
apotterri: The changes in this commit address this concern: `test_simple.py` now imports `appmap.unittest`, and the test cases confirm that the results are correct when driven by `pytest`, `appmap.unittest`, as well as `unittest`.

apotterri: The changes in da92d08b243a40ab2033233a4fa9dcf3a30221f6 address this concern: `test_simple.py` now imports `appmap.unittest`, and the test cases confirm that the results are correct when driven by `pytest`, `appmap.unittest`, as well as `unittest`.

apotterri:
@dividedmind this is ready for another look. I've kept the commits related to unittest support separate, because I think they'll be easier to review this way.  LMK if you think they should be squashed (or not) before merging.

dividedmind (approved):
Looks much better, thanks!

Issue #79, opened 2021-04-14
Misago failing tests


ptrdvrk:
Misago tests are failing with the `free-functions_20210310` branch client.

Failing tests:
- misago/users/tests/test_twitter_profilefield.py
- misago/users/tests/test_user_editdetails_api.py
- misago/core/tests/test_chunk_queryset.py
- misago/themes/admin/tests/test_creating_and_editing_themes.py
- misago/users/tests/test_bio_profilefield.py
- misago/users/tests/test_gender_profilefield.py

appmap.yml:
```yaml
name: misage
packages:
  - path: misago
```

[misago_test_errors.log](https://github.com/applandinc/appmap-python/files/6313505/misago_test_errors.log)

dividedmind:
I'm not getting any failures anymore with #86 and #87.

dividedmind:
Fixed by #86 and #87.

PR #86, opened 2021-04-24
Use repr() by default for stringification

dividedmind:
`repr()` is usually the preferred programmer-friendly way to display an object. `str()` is a user representation (for many different meanings of a user), and as such can sometimes cause an expensive and state-changing stringification.

As an example for how that might be harmful, consider `misago.users.admin.forms.EditUserForm.Meta.__init__`:
https://github.com/rafalp/Misago/blob/da14e8ef712a77603fdfec8bbff67c526c7d0c5c/misago/users/admin/forms.py#L202

This method first delegates to superclass, then calls `profilefields.add_fields_to_admin_form(self.request, self.instance, self)`; `self` here is the form. If the appmap hook tries to use `str()` for stringification of this `form` parameter, this will cause HTML rendering and, before that, form validation. Note this happens before the form is completely constructed; some fields have not been added yet. Consequently a validation error in one of these fields will be missed (as the form had validated itself, even though it wasn't completely constructed); this leads to test failure in https://github.com/rafalp/Misago/blob/da14e8ef712a77603fdfec8bbff67c526c7d0c5c/misago/users/tests/test_twitter_profilefield.py#L51

(I'll also note that several pagefuls of HTML aren't particularly useful when reading an appmap. `<UserFormFinal bound=True, valid=Unknown, fields=([...])>` (ellipsis mine) provided by `repr()` is much more informative.)

Generally, consider that we're often trying to record the values inside method calls; however, inside a method call, any invariants the class might have (and `str()` might rely upon) cannot be guaranteed to hold. `repr()` shouldn't suffer from this issue, as it's routinely used by REPL and, in particular, debuggers, to get the object representation in code paused in embarassing places, and so a reasonable programmer would implement it with as few assumptions (and overhead!) as possible, including the assumption of class invariants being upheld.

(Related to #79.)

### DIFF
--- a/appmap/_implementation/event.py
+++ b/appmap/_implementation/event.py
@@ -53,18 +53,15 @@ class _EventIds:
 
 def display_string(val):
     # If we're asked to display parameters, make a best-effort attempt
-    # to get a string value for the parameter using str() and
-    # repr(). If parameter display is disabled, or str() and repr()
-    # both raised, just formulate a value from the class and id.
+    # to get a string value for the parameter using repr(). If parameter
+    # display is disabled, or repr() has raised, just formulate a value
+    # from the class and id.
     value = None
     if Env.current.display_params:
         try:
-            value = str(val)
+            value = repr(val)
         except Exception:  # pylint: disable=broad-except
-            try:
-                value = repr(val)
-            except Exception:  # pylint: disable=broad-except
-                pass
+            pass
 
     if value is None:
         class_name = fqname(type(val))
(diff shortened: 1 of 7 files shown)


appmap/_implementation/event.py
```diff
@@ -59,10 +59,10 @@ def display_string(val):
     value = None
     if Env.current.display_params:
         try:
-            value = str(val)
+            value = repr(val)
         except Exception:  # pylint: disable=broad-except
             try:
-                value = repr(val)
+                value = str(val)
```
apotterri, line 65: I wonder if we even need this? If `repr` didn't work, what are the chances that `str` will?

apotterri (approved):
One comment to consider, but this looks good to me.

I've wondered for a while whether we should be using `repr`. While switching will certainly improve our robustness somewhat, it's definitely not a panacea. I've seen many classes that simply have `repr` as an alias for `str`, or vice versa.

appmap/_implementation/event.py
dividedmind: Good point, I'll remove it.

PR #87, opened 2021-04-25
Avoid capturing queries caused by calling repr()

dividedmind:
Calling repr() on QuerySet objects in django causes a database query. We don't want these queries to end up in the appmap, since they're not part of the application flow.

Moreover, some tests might record and examine queries using CursorDebugWrapper; we don't want them to see those queries, either.

For an example error caused by this, see https://github.com/rafalp/Misago/blob/da14e8ef712a77603fdfec8bbff67c526c7d0c5c/misago/core/tests/test_chunk_queryset.py#L29 -- this tests counts queries and it fails if it sees additional query caused by us calling repr() on the QuerySet.

Related to #79 .


kgilpin:
This seems like a good idea to me. But I would probably need Alan’s ok
also.

apotterri (approved):
LGTM, thanks.

PR #88, opened 2021-04-26
Add SQLAlchemy integration

### LINKED ISSUE #64: SQLAlchemy integration captures sql_query events

dividedmind:
SQL queries are captured as long as `appmap.sqlalchemy` is imported before connections are made. Flask integration tries to import it by default, since both are often used in conjunction.

Note captured SQL is parametrized, for example `INSERT INTO users (name, fullname) VALUES (?, ?)`; this is because of how SQLAlchemy executes SQL -- it's actually sent parametrized to the server, the full substituted SQL is never produced outside the database driver (as far as I can tell). This is in line with SQLAlchemy own logging; it adds another line with values:
```
INFO:sqlalchemy.engine.Engine:INSERT INTO users (name, fullname) VALUES (?, ?)
INFO:sqlalchemy.engine.Engine:[generated in 0.00047s] ('jack', 'Jack Jones')
```

Not sure if we should do the same; it can be added to the SQL as a comment, or we could add this as call parameters.

Fixes #64 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   recording them in the `message` attribute of the `call` event.
 - [#83] Capture HTTP request headers in django and flask.
 - [#53] Module-scoped functions are now recorded.
+- [#64] Capture SQL queries in SQLAlchemy.
 ### Changed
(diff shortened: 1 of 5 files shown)


apotterri (approved):
Looks good, thanks very much.

At Kevin's suggestion, I added #92. We can consider capturing query parameters as a follow on.

PR #93, opened 2021-04-28
Http response headers

### LINKED ISSUE #90: Capture HTTP response headers

dividedmind:
Fixes #90

### DIFF
--- a/appmap/_implementation/event.py
+++ b/appmap/_implementation/event.py
@@ -252,32 +252,39 @@ class MessageEvent(Event):
             self.message.append(message_object)
 
 
+def filter_http_headers(headers):
+    """Filter out HTTP headers which shouldn't go into appmap events.
+    Return None if none left."""
+    headers = {
+        k: v for k, v in headers.items()
+        if k not in ['Content-Type', 'Authorization', 'Host', 'User-Agent']
+    }
+    return headers if len(headers) > 0 else None
+
+
 class HttpRequestEvent(MessageEvent):
+    """A call AppMap event representing an HTTP request."""
     __slots__ = ['http_server_request']
 
     def __init__(self, request_method, path_info, message_parameters,
                  normalized_path_info=None, protocol=None, headers=None):
         super().__init__(message_parameters)
 
-        if headers is not None:
-            mime_type = headers.get('Content-Type')
-            authorization = headers.get('Authorization')
-            headers = {
-                k: v for k, v in headers.items()
-                if k not in ['Content-Type', 'Authorization', 'Host', 'User-Agent']
-            }
-
-        http_server_request = {
+        request = {
             'request_method': request_method,
             'protocol': protocol,
             'path_info': path_info,
             'normalized_path_info': normalized_path_info,
-            'mime_type': mime_type,
-            'authorization': authorization,
-            'headers': headers if headers is not None and len(headers) > 0 else None
         }
 
-        self.http_server_request = compact_dict(http_server_request)
+        if headers is not None:
+            request.update({
+                'mime_type': headers.get('Content-Type'),
+                'authorization': headers.get('Authorization'),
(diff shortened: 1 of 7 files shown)


dividedmind:
Dang, another special handling for 3.5 needed... #94

appmap/_implementation/event.py
```diff
@@ -249,32 +250,39 @@ def __init__(self, message_parameters):
             self.message.append(message_object)
 
 
+def filter_http_headers(headers):
+    """Filter out HTTP headers which shouldn't go into appmap events.
+    Return None if none left."""
+    headers = {
+        k: v for k, v in headers.items()
+        if k not in ['Content-Type', 'Authorization', 'Host', 'User-Agent']
```
kgilpin, line 258: Shall we remove this special case on Host and User-Agent, and include them?

kgilpin: If we just write all the headers, it will make it easier to remove those `mime_type`, `authorization` parameters that you don't like :-) https://github.com/applandinc/appmap-ruby/pull/117/files#diff-289cc4c1e9edad7c0982425150a91c7d8df3c0bbeeb8ec5d5a55073c3bad537fR86

appmap/_implementation/event.py
```diff
@@ -297,13 +305,21 @@ def __init__(self, parent_id, elapsed, return_value):
...
 
-    def __init__(self, status_code, mime_type, **kwargs):
+    def __init__(self, status_code, headers=None, **kwargs):
         super().__init__(**kwargs)
-        self.http_server_response = {
-            'status_code': status_code,
-            'mime_type': mime_type
+
+        response = {
+            'status_code': status_code
```
kgilpin, line 312: FYI, the same PR I linked above fixes `status` -> `status_code` in Ruby.

appmap/_implementation/event.py
```diff
@@ -249,32 +250,39 @@ def __init__(self, message_parameters):
...
+    """Filter out HTTP headers which shouldn't go into appmap events.
+    Return None if none left."""
+    headers = {
+        k: v for k, v in headers.items()
+        if k not in ['Content-Type', 'Authorization', 'Host', 'User-Agent']
+    }
+    return headers if len(headers) > 0 else None
+
+
 class HttpRequestEvent(MessageEvent):
```
kgilpin, line 263: By the way, we don't have real message parameters for Flask - https://github.com/applandinc/appmap-python/blob/897fd4ed609c23ee3f0c76c23747dc1db1a0729b/appmap/flask.py#L70

Although we do for Django - https://github.com/applandinc/appmap-python/blob/be049598c95d7536685201dc09f3be79ae50d12a/appmap/django.py#L74

Without message parameters, Swagger gen won't work too well.

kgilpin (approved):
See comments.

appmap/_implementation/event.py
dividedmind: > Shall we remove this special case on Host and User-Agent, and include them?

Now you're telling me!

> If we just write all the headers, it will make it easier to remove those `mime_type`, `authorization` parameters that you don't like :-) https://github.com/applandinc/appmap-ruby/pull/117/files#diff-289cc4c1e9edad7c0982425150a91c7d8df3c0bbeeb8ec5d5a55073c3bad537fR86

Yeah, that's what I propose at https://github.com/applandinc/appmap/pull/15

appmap/_implementation/event.py
dividedmind: I didn't even notice it was broken...

appmap/_implementation/event.py
kgilpin: I know that's why I'm suggesting it.

apotterri (approved):
LGTM

PR #95, opened 2021-05-04
Capture HTTP client request and response

dividedmind:


appmap/test/test_http.py
```diff
@@ -0,0 +1,32 @@
...
+    assert (message[1].items() >= { 'name': 'q2', 'value': "'🦠'" }.items()) \
+        or (message[1].items() >= { 'name': 'q2', 'value': "'\\U0001f9a0'" }.items())
+
+    assert events[1].http_client_response.items() >= {
+        'status_code': 200,
+        'mime_type': 'text/plain; charset=utf-8'
+    }.items()
+
+@pytest.yield_fixture
+def mock_requests():
```
apotterri, line 29: Apparently, this can just be `@pytest.fixture` now: https://docs.pytest.org/en/6.2.x/yieldfixture.html .

Also, you might want to use the `name` argument to `fixture`, and get rid of `pylint: disable=redefined-outer-name`: https://docs.pytest.org/en/6.2.x/reference.html#pytest-fixture .

apotterri (approved):

PR #96, opened 2021-05-04
Add a missing return

apotterri:
When using `pytest` to drive tests, the decorator for test cases wasn't
returning the value of the wrapped function.


kgilpin (approved):

dividedmind:
Argh, I was looking at that code and I thought something didn't feel right. Shouldn't we add a test that would have caught this?

apotterri:
The original implementation wasn't actually wrong, for vanilla `unittest` test cases, the return value from a test method is ignored: https://github.com/python/cpython/blob/3.9/Lib/unittest/case.py#L550 .

scrapy uses `twisted.trial` for testing, though, and it *does* examine the return value: https://twistedmatrix.com/documents/current/core/howto/trial.html#creating-and-testing-the-client .

I've tried to write a test using `twisted.trial` that fails if the return value is missing. This is complicated by this behavior mentioned in the `twisted.trial` doc:
> When a test returns a Deferred, the reactor is run until the Deferred fires and its callbacks run. The important thing to do here is to not forget to return the Deferred . If you do, your tests will pass even if nothing is asserted.

Writing a test using `twisted.trial` isn't trivial, and the framework is pretty heavy, so this doesn't seem like a great way to go, at least for now. I'll look for another way to test this change that doesn't required it.

apotterri:
@dividedmind So, I couldn't come up with a lightweight way to test this, but I did find a way using `twisted.trial`. Will you take another look, please?

dividedmind (approved):
Good job reorganizing the tests, thanks!

PR #98, opened 2021-05-06
Add recording to Django integration

### LINKED ISSUE #3: Add django integration for remote recording

dividedmind:
Fixes #3.


kgilpin:
@dustinbyrne made a suggestion that we have a "safety" on remote recording, so that people don't accidentally run it in production. So, in appmap-ruby, you have to set the environment variable `APPMAP_RECORDING=true` in order to activate the remote recording routes. This way, people are unlikely to accidentally open up access to recording of their production apps.

kgilpin (changes requested):
Can you add a "safety" flag to this? See my comment in the PR.

dividedmind:
> Can you add a "safety" flag to this? See my comment in the PR.

It needs `APPMAP=true`, see https://github.com/applandinc/appmap-python/blob/django-recording/README.md#run-your-web-app and test at https://github.com/applandinc/appmap-python/blob/3e0a623d1937fa343d509ed4d3b76b01aded93bb/appmap/test/web_framework.py#L61

appmap/_implementation/generation.py
```diff
@@ -114,5 +115,8 @@ def default(self, o):
 
 
 def dump(recording, metadata=None, indent=None):
+    """Generate a JSON representation of an appmap."""
+    if hasattr(recording, 'events'):
```
apotterri, line 119: Changing `appmap` to take `events` instead of `recording` makes sense to me. It seems to me that the corresponding change should be made here: change `recording` to `events`, and get rid of the `hasattr` check. I don't think we need to worry about the fact that this changes the interface.

appmap/test/normalize.py
```diff
@@ -46,6 +46,13 @@ def normalize_metadata(metadata):
                 v = f.pop('version')
                 assert v == pytest.__version__
 
+def normalize_headers(dct):
+    """Remove some headers which are variable between implementations.
+    This allows sharing tests between web frameworks, for example.
+    """
+    for hdr in ['User-Agent', 'Content-Length', 'ETag', 'Cookie', 'Host']:
+        dct.pop(hdr, None)
```
apotterri, line 54: I think it's worth confirming that the values of these keys are of the correct type, as we're doing for `elapsed` and `object_id` below

appmap/test/test_django.py
```diff
@@ -1,25 +1,18 @@
...
+import pytest
+
+
 import appmap.django  # noqa: F401
 
+from .web_framework import TestRequestCapture, TestRecording
 
-# Note that with django configured this way, django.test.Client
-# requests for '/test' will print warning messages that look like
-# `WARNING:django.request:Not Found: /test`. These don't appear to
```
apotterri, line 16: You moved this code, but not the comment. Have you made other changes that got rid of these warnings (when you run `pytest -svv appmap`? If not, can you put the comment (or some variation of it) back, please?

apotterri (changes requested):

appmap/test/test_django.py
dividedmind: Yes, I added the views and routes.

appmap/_implementation/generation.py
dividedmind: It's a public interface and I don't think there's a good reason to break it.

dividedmind:
Note: I've also added a change that should prevent problems like https://travis-ci.com/github/applandinc/appmap-python/builds/225454355#L1086:
> docker: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit.

appmap/_implementation/generation.py
apotterri: I doubt there's anybody using it at this point. I think we should fix it before others start to use it.

apotterri:
> Note: I've also added a change that should prevent problems like https://travis-ci.com/github/applandinc/appmap-python/builds/225454355#L1086:
> 
> > docker: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit.

I fixed this by updating the build config to authenticate with docker hub.

I'd really like to keep the step of installing in a docker container, to best simulate installing into a pristine environment.

appmap/_implementation/generation.py
dividedmind: I also think it would make the interface worse, cf.
```python
rec = appmap.Recording()
with rec:
  ...
print(appmap.generation.dump(rec))
```
vs.
```python
print(appmap.generation.dump(rec.events))
```

kgilpin:
> So, I guess the only other question here is whether requiring `APPMAP=true` is enough protection, or whether `APPMAP_RECORDING=true` should also be required.
> 
> @kgilpin What would you like to see here?

It seems to me like having `APPMAP=true` is enough protection, certainly for now. Unlikely that someone would have that setting accidentally enabled in production.

apotterri (commented):
It looks like there are still a couple of changes to make here: https://github.com/applandinc/appmap-python/pull/98/files#r630042113 and https://github.com/applandinc/appmap-python/pull/98#issuecomment-836752496 .

Once those are finished up, this will be good to go. Thanks!

appmap/_implementation/generation.py
```diff
@@ -114,5 +115,8 @@ def default(self, o):
...
+        recording = recording.events
```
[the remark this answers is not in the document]
dividedmind: I can't, this will break https://github.com/applandinc/appmap-python/pull/98/files#diff-06d67252592021259f8167f0ef2fc8970192a7aa947478e8d00d210e19067e3aR162 -- this is exactly why I've made this change.

dividedmind:
> [#98 (comment)](https://github.com/applandinc/appmap-python/pull/98#issuecomment-836752496) .

I actually meant to do this while rebasing for the merge, but ok, I'll do it now.

apotterri:
> > [#98 (comment)](https://github.com/applandinc/appmap-python/pull/98#issuecomment-836752496) .
> 
> I actually meant to do this while rebasing for the merge, but ok, I'll do it now.

Doing it while rebasing is fine, just didn't want it to get forgotten.

appmap/_implementation/generation.py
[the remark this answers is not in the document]
apotterri: I'm confused. Doesn't that line you linked to return a recording that has an `events` attribute? When is `dump` called with with an object that doesn't have an `events` attribute?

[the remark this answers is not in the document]
dividedmind: No, it returns event list directly, cf. https://github.com/applandinc/appmap-python/blob/879ee909fcb60b0d7527fc86d62500a90b627c81/appmap/_implementation/recording.py#L232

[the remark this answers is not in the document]
apotterri: I still think the control coupling in this interface is bad, but I'm tired of fighting about it.

apotterri (approved):

kgilpin (approved):

PR #100, opened 2021-05-07
Write recordings atomically

### LINKED ISSUE #99: Recordings are written atomically
Instead of opening an `appmap.json` file and writing its contents, the recording should be written to a temporary file, then renamed.

apotterri:
When writing a recording, save it to a temp file first, then rename it
to the final file.

Fixes #99.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#90] Capture HTTP response headers in django and flask.
 
 ### Changed
+- A recording is now created atomically. It is first written to a temp file, which is then
+  renamed to the final file.
 - Headers such as Host, User-Agent, Authorization and Content-Type are no longer
   filtered out in HTTP event `headers` field.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,13 +5,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- A recording is now created atomically. It is first written to a temp file, which is then
+  renamed to the final file.
 ## [0.10.0] - 2021-05-07
 ### Added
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#90] Capture HTTP response headers in django and flask.
 
 ### Changed
+- A recording is now created atomically. It is first written to a temp file, which is then
+  renamed to the final file.
```
dividedmind, line 14: This should go under *unreleased* I think. 
(Git doesn't do a great job merging changelogs, it's good to keep an eye on these.)

appmap/_implementation/testing_framework.py
```diff
@@ -124,4 +126,11 @@ def record(self, klass, method, **kwds):
...
         metadata = fi.metadata
         metadata.update(self.metadata)
-        (self.appmap_path / filename).write_text(generation.dump(rec, metadata))
+        appmap_json = self.appmap_path / filename
+        with NamedTemporaryFile(mode='w', dir=self.appmap_path, delete=False) as f:
+            f.write(generation.dump(rec, metadata))
+            try:
+                os.remove(appmap_json)
+            except FileNotFoundError:
+                pass
```
dividedmind, line 135: This is not necessary; `os.replace()` replaces the destination (as the name implies). What's more, this can lead to race with eg. an editor that has the file open. It could see the file removed and erronously inform the user it has been removed, instead of replaced. `os.replace()` replaces atomically if possible.

appmap/_implementation/testing_framework.py
```diff
@@ -124,4 +126,11 @@ def record(self, klass, method, **kwds):
...
+            os.replace(f.name, appmap_json)
```
dividedmind, line 136: This is not guaranteed to work correctly on every operating system, since the file is still open at this point. Either close it first explicitly or move the move out of the context.

dividedmind (changes requested):

CHANGELOG.md
apotterri: Oops, yeah. Fixed.

PR #103, opened 2021-05-12
Capture function comments in the classMap

### LINKED ISSUE #102: Function docstrings are captured as comments in a recording
If there's no docstring, use the function's comment instead.

apotterri:
Fixes #102.

Also, fix some warnings and failures caused by the release werkzeug 2.0.0.


apotterri:
@dividedmind you'll probably want to pull these changes: https://github.com/applandinc/appmap-python/pull/103/commits/6a86375b213d866a68a84d2b60e44a7c4386717e

dividedmind:
> @dividedmind you'll probably want to pull these changes: [6a86375](https://github.com/applandinc/appmap-python/commit/6a86375b213d866a68a84d2b60e44a7c4386717e)

These are taken care of by 92c2331 and e906f45 which I think are cleaner. Thanks though :)

appmap/_implementation/event.py
```diff
@@ -210,13 +211,49 @@ def set_params(params, instance, args, kwargs):
...
+    @lru_cache(maxsize=None)
+    def lineno(self):
+        return self.function_location[1]
+
+    @property
+    @lru_cache(maxsize=None)
+    def comment(self):
+        comment = inspect.getdoc(self._fn)
+        if comment is not None:
+            comment = '\n'.join([f'# {s}' for s in comment.splitlines()])
```
dividedmind, line 249: I don't think it makes sense to prepend the `#`. The spec certainly doesn't require it (and if it did, it would be really weird with languages that use other comment markers.) I'd just leave the doc as is.

appmap/test/data/expected.appmap.json
```diff
@@ -324,6 +324,7 @@
         {
           "name": "dump",
           "type": "function",
+          "comment": "# Serialize a Python object into a YAML stream.\n# If stream is None, return the produced string instead.",
```
dividedmind, line 327: This seems too brittle to me. (What if the yaml package changed their doc?) Can we filter this instead?

dividedmind (changes requested):

appmap/_implementation/event.py
apotterri: The spec doesn't require it, but appmap-swagger does: https://github.com/applandinc/appmap-swagger/blob/main/method.js#L82 . And, the comments emitted by appmap-ruby have the leading `# `.

appmap/test/data/expected.appmap.json
apotterri: Good point, I'll do that.

appmap/_implementation/event.py
dividedmind: This seems like a problem in appmap-swagger. Can we add a comment here to signal the workaround and open an issue there? It's going to be a problem with other languages.

(Appmap-ruby works like that because it just cuts the code verbatim (sans indentation). It doesn't add the hashes :) But actually it would make sense to remove them universally, once appmap-swagger can handle that.)

apotterri: Yes, I agree that it's a problem in appmap-swagger. I [fixed it](https://github.com/applandinc/appmap-swagger/pull/7), so I'll update this.

dividedmind: I think you forgot to remove this after all.

dividedmind (changes requested):

apotterri:
https://github.com/applandinc/appmap-python/pull/103#discussion_r634607531

Thanks, fixed now.

dividedmind (approved):

PR #106, opened 2021-05-13
Capture test results

### LINKED ISSUE #101: Record test case result
Per applandinc/appmap#17

dividedmind:
Fixes #101.

Note e906f45 is an unrelated change adapted from #98, but required due to failing tests with a new flask version.

Also note this is blocked by https://github.com/applandinc/appmap/pull/17


appmap/pytest.py
```diff
@@ -59,5 +60,10 @@ def pytest_pyfunc_call(pyfuncitem):
...
                 method_id=pyfuncitem.originalname,
-                location=pyfuncitem.location):
-            yield
+                location=pyfuncitem.location) as metadata:
+            result = yield
+            try:
+                with testing_framework.collect_result_metadata(metadata):
+                    result.get_result()
+            except:  # pylint: disable=bare-except
+                pass  # exception got recorded in metadata
```
apotterri, line 69: Shouldn't this be allowing the exception to propagate?

appmap/unittest.py
```diff
@@ -34,8 +34,8 @@ def _args(test_case, *_, isTest=False, **__):
     with session.record(
             test_case.__class__,
             method_name,
-            location=location):
-        with wrapped(*args, **kwargs):
+            location=location) as metadata:
+        with wrapped(*args, **kwargs), testing_framework.collect_result_metadata(metadata):
```
apotterri, line 38: Nice, I didn't know about this way of using multiple context managers.

appmap/test/normalize.py
```diff
@@ -46,6 +46,13 @@ def normalize_metadata(metadata):
                 v = f.pop('version')
                 assert v == pytest.__version__
 
+def normalize_headers(dct):
+    """Remove some headers which are variable between implementations.
+    This allows sharing tests between web frameworks, for example.
+    """
+    for hdr in ['User-Agent', 'Content-Length', 'ETag', 'Cookie', 'Host']:
+        dct.pop(hdr, None)
```
apotterri, line 54: You were going to add a check for the type of the value here, weren't you?

apotterri (changes requested):
A question, and one minor changes. Looks good otherwise.

appmap/pytest.py
dividedmind: No, the exception is captured in the `result`. Hookwrappers shouldn't throw (unless there's a bug).

appmap/test/normalize.py
dividedmind: Right, it's in the other PR: 54f64eebe64854a25d623774509306d0f68bd560; I cherry-picked an older version here unintentionally. While merging the newer version will prevail :)

apotterri (approved):

PR #110, opened 2021-05-22
Capture message parameters in flask

### LINKED ISSUE #108: Capture message parameters in flask

dividedmind:
Fixes #108.

### DIFF
--- a/appmap/test/test_flask.py
+++ b/appmap/test/test_flask.py
@@ -20,3 +20,23 @@ def flask_client(data_dir, monkeypatch):
 
     with app.app.test_client() as client:  # pylint: disable=no-member
         yield client
+
+
+@pytest.mark.appmap_enabled
+def test_message_path_segments(events, client):
+    client.get('/post/alice/42/summary')
+
+    assert events[0].message == [
+        {
+            'name': 'username',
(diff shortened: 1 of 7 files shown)


appmap/test/test_flask.py
```diff
@@ -40,3 +40,15 @@ def test_message_path_segments(events, client):
...
+@pytest.mark.appmap_enabled
+@pytest.mark.parametrize('url,expected', [
+    ('/user/test_user', '/user/{username}'),
+    ('/post/123', '/post/{post_id}'),
+    ('/post/test_user/123/summary', '/post/{username}/{post_id}/summary')
+])
+def test_path_normalization(client, events, url, expected):
+    client.get(url)
+    normalized = events[0].http_server_request['normalized_path_info']
+    assert normalized == expected
```
apotterri, line 54: I figured I'd put this back when django supported it (hence the comment in web_framework.py). Is there some reason it needs to be here?

apotterri (approved):

appmap/test/test_flask.py
dividedmind: Not really, I just noticed it while rebasing and figured might as well put it in flask test for now instead of having dead code.

PR #111, opened 2021-05-24
Limit appmap file name to 255 characters

### LINKED ISSUE #91: AppMap filenames must be shorter than 255 characters
appmap-python should ensure that names of AppMaps are shorter than 255 characters to avoid issues like https://github.com/applandinc/appmap-java/issues/75.

dividedmind:
Fixes #91

### DIFF
--- a/appmap/_implementation/env.py
+++ b/appmap/_implementation/env.py
@@ -1,11 +1,12 @@
 """Initialize from the environment"""
+
+from pathlib import Path
 import logging
 import logging.config
-import os
-import os.path
+from os import environ
 
-_cwd = os.getcwd()
-_bootenv = os.environ.copy()
+_cwd = Path.cwd()
+_bootenv = environ.copy()
 
 class _EnvMeta(type):
     def __init__(cls, *args, **kwargs):
@@ -33,12 +34,11 @@ class Env(metaclass=_EnvMeta):
         if env:
             self._env.update(env)
 
-        self._root_dir = str(os.path.join(self._cwd) + '/')
+        self._root_dir = str(self._cwd) + '/'
         self._root_dir_len = len(self._root_dir)
 
-        output_dir = self.get("APPMAP_OUTPUT_DIR",
-                               os.path.join('tmp', 'appmap'))
-        self._output_dir = os.path.abspath(output_dir)
+        output_dir = Path(self.get('APPMAP_OUTPUT_DIR', 'tmp/appmap'))
+        self._output_dir = output_dir.resolve()
 
         self._configure_logging()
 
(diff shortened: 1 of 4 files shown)


appmap/_implementation/env.py
```diff
@@ -38,7 +39,7 @@ def __init__(self, env=None):
 
         output_dir = self.get("APPMAP_OUTPUT_DIR",
                                os.path.join('tmp', 'appmap'))
-        self._output_dir = os.path.abspath(output_dir)
+        self._output_dir = Path(os.path.abspath(output_dir))
```
apotterri, line 42: Can this use `Path.resolve` instead?

appmap/_implementation/testing_framework.py
```diff
@@ -90,11 +92,40 @@ def metadata(self):
...
+        part = NAME_MAX - len(APPMAP_SUFFIX) - 1 - HASH_LEN
+        basename = basename[:part] + '-' + name_hash(basename[part:])[:HASH_LEN]
+    filename = basename + APPMAP_SUFFIX
+
+    if not basedir.exists():
+        basedir.mkdir(parents=True, exist_ok=True)
+
+    with NamedTemporaryFile(mode='w', dir=basedir, delete=False) as tmp:
+        tmp.write(contents)
+    os.replace(tmp.name, basedir / filename)
```
apotterri, line 122: I wonder if we also need to worry about a failure here because the total path length exceeds the filesystem's limit? IIRC, we had a user report a failure when trying to write an appmap to a mounted filesystem, and it seemed like the root cause was the path length. I can't find the message now, though, so maybe I'm misremembering....

apotterri (approved):
LGTM. I made a couple of suggestions, but I don't feel strongly about them.

appmap/_implementation/testing_framework.py
dividedmind: I suppose this is an issue a user could run into, `PATH_MAX` is 4096 by default on Linux. But I don't think it makes sense for us to handle this; there's not much we can do, if a user has such a deep directory hierarchy then they're bound to run into problems sooner or later anyway. The error message in this case should be descriptive enough (I haven't checked though). And I think it's more important to have predictable names. Tl;dr I don't think should be in our purview.

appmap/_implementation/env.py
dividedmind: Sure, good point.

dividedmind: I've split env cleanup into a separate commit and added a couple more changes, do you want to take a look? https://github.com/applandinc/appmap-python/pull/111/commits/ede8e19fb5c5cfe2bd5eedd67c35d5cd65916d21

apotterri: Looks good, I'd say this is ready to go.

PR #112, opened 2021-05-25
Show normalized_path_info for django

apotterri:
Add normalized_path_info to http_server_request events captured by the
django integration.


dividedmind (changes requested):
Path arguments should be captured in `message`, cf. https://github.com/applandinc/appmap-python/blob/42a7f1e4bde80c5fe0d1e087efbf1d93b9377c5a/appmap/test/test_flask.py#L26 -- that test should also be moved into *appmap.test.web_framework*.

dividedmind (approved):
Thanks!

Issue #70, opened 2021-03-26
Exception in handling django HTTP request causes ill-formed AppMap


apotterri:
`appmap.django.Middleware.__call__` doesn't handle exceptions. So, if `get_response` raises, there won't be a "return" event added to match the `http_server_request` "call" event.

dividedmind:
I'm not sure what to review. Is there a pull request? Please link it.

PR #113, opened 2021-05-25
Added exception handler for request in the Middleware class of django.

snowmaksym:
Fixes: #70

### DIFF
--- a/appmap/django.py
+++ b/appmap/django.py
@@ -18,10 +18,12 @@ from django.urls import get_resolver, resolve
 import time
 import json
 import re
+import sys
 
 from appmap._implementation import generation
 from appmap._implementation.env import Env
 from appmap._implementation.event import (
+    ExceptionEvent,
     HttpServerRequestEvent,
     HttpServerResponseEvent,
     ReturnEvent,
@@ -206,7 +208,18 @@ class Middleware:
             )
             self.recorder.add_event(call_event)
 
-        response = self.get_response(request)
+        try:
+            response = self.get_response(request)
+        except:
+            if self.recorder.enabled:
+                duration = time.monotonic() - start
+                exception_event = ExceptionEvent(
+                    parent_id=call_event.id,
+                    elapsed=duration,
+                    exc_info=sys.exc_info()
+                )
+                self.recorder.add_event(exception_event)
+            raise
 
         if self.recorder.enabled:
             duration = time.monotonic() - start
--- a/appmap/test/test_django.py
+++ b/appmap/test/test_django.py
@@ -5,6 +5,7 @@ from pathlib import Path
 
 import django
 import django.conf
+import django.core.handlers.exception
 import django.db
 import django.http
 from django.template.loader import render_to_string
@@ -114,6 +115,9 @@ def user_post_view(_request, username, post_id):
 
 def echo_view(request):
     return django.http.HttpResponse(request.body)
+    
+def exception_view(_request):
+    raise RuntimeError("An error")
 
 def user_included_view(_request, username):
     return django.http.HttpResponse(f'user {username}, included')
@@ -126,6 +130,7 @@ urlpatterns = [
     path('post/<username>/<int:post_id>/summary', user_post_view),
     re_path(r'^post/unnamed/(\d+)$', post_unnamed_view),
     path('echo', echo_view),
+    path('exception', exception_view),
     re_path(r'^post/included/', include([
             path('<username>', user_view)]))
 ]
@@ -149,6 +154,27 @@ def test_included_view(client, events):
         'path_info': '/post/included/test_user',
         'normalized_path_info': '/post/included/{username}'
     })
+
+@pytest.mark.appmap_enabled
+def test_exception(client, events, monkeypatch):
+    def raise_on_call(*args):
+        raise RuntimeError('An error')
+    monkeypatch.setattr(django.core.handlers.exception, 'response_for_exception', raise_on_call)
+
+    with pytest.raises(RuntimeError):
+        client.get('/exception')
+
+    assert events[0].http_server_request == DictIncluding({
+        'request_method': 'GET',
+        'path_info': '/exception',
+        'protocol': 'HTTP/1.1'
+    })
+
+    assert events[1].parent_id == events[0].id
+    assert events[1].exceptions == [DictIncluding({
+            'class': 'builtins.RuntimeError',
+            'message': 'An error'
+    })]
 
 django.conf.settings.configure(
     DATABASES={'default': {'ENGINE': 'django.db.backends.sqlite3', 'NAME': ':memory:'}},
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#119] Record template rendering in Django and flask.
 
 ### Fixed
+- [#70] Django integration now records an ExceptionEvent when an unhandled
+  exception is raised within Django itself while processing a request.
 - When an argument to a method is missing, don't raise an exception in the
   appmap code. Instead omit the missing parameter and allow the original
   function call to raise ArgumentError if appropriate.


apotterri (changes requested):
It's hard to tell whether this actually addresses #70. Please add at least one test that shows that if `get_response` raises, the generated appmap matches an expected appmap.

appmap/django.py
```diff
@@ -143,8 +150,42 @@ def __call__(self, request):
...
+                parent_id=self.call_event_id,
+                elapsed=duration,
+                exc_info=sys.exc_info()
+            )
+            self.recorder.add_event(exception_event)
+
+            self.call_event_id = None
+            self.start = 0
+
+            return HttpResponse("Error processing the request.", status=500)
```
dividedmind, line 186: What if the app wants to return a different response for the exception?

dividedmind (changes requested):
Like Alan said, this needs a test. Also, wouldn't it be easier to simply add a `try` in the middleware instead? 

BTW, it's not clear to me when an exception might be raised in `get_response` — surely if the application code raises an exception, it will have been handled and turned into the correct response before it gets to us? @apotterri, do you remember if there was a specific scenario where it lead to broken appmaps, or was the intention to just be defensive?

apotterri:
I agree with @dividedmind: I think it would be better to wrap `get_response` in a `try` block. It should create the `Exception` event and re-raise the exception.

An unhandled exception in a view will propagate through `get_response`. If we only ever mapped bug-free applications, we would be able to rely on their exception handling. :) It seems better to be defensive, and ensure that a correctly formed appmap is always generate.

snowmaksym:
I've deleted `process_exception` function in `Middleware` class, and wrapped  `get_response` in a try block.
But I'm not sure if there is an exception when `get_response` function is called.

appmap/django.py
```diff
@@ -131,7 +132,18 @@ def __call__(self, request):
             )
             self.recorder.add_event(call_event)
 
-        response = self.get_response(request)
+        try:
+            response = self.get_response(request)
+        except Exception as e:
```
dividedmind, line 137: You're not using `e`; might as well not bind it and just reraise with plain `raise`.

dividedmind (changes requested):
This looks better now. Still needs a test though.

apotterri:
> I've deleted `process_exception` function in `Middleware` class, and wrapped `get_response` in a try block.
> But I'm not sure if there is an exception when `get_response` function is called.

The way to be sure is to arrange for an exception to get raised from a view. To get started, I'd suggest adding a new route and its corresponding view to `test_django.py`. Then, add a test that requests the new route and verifies that the appropriate events are generated.

As a follow on task, we should move the new test into `web_framework.py` and add the appropriate function to `test_flask.py`.

appmap/test/test_django.py
```diff
@@ -107,6 +111,7 @@ def echo_view(request):
     django.urls.path('post/<username>/<int:post_id>/summary', user_post_view),
     django.urls.re_path(r'^post/unnamed/(\d+)$', post_unnamed_view),
     django.urls.path('echo', echo_view),
+    django.urls.re_path('exception', exception_view)
```
apotterri, line 114: This can just be `path`, rather than `re_path`.

appmap/test/test_django.py
```diff
@@ -116,6 +121,21 @@ def test_unnamed(client, events):
     # message attribute.
     assert len(events[0].message) == 0
 
+def test_exception(client, events):
+    try:
```
apotterri, line 125: You should use `pytest.raises` to verify that the expected exception is being raised: https://docs.pytest.org/en/reorganize-docs/new-docs/user/pytest_raises.html .

appmap/test/test_django.py
```diff
@@ -116,6 +121,21 @@ def test_unnamed(client, events):
...
+        client.get('/exception')
+
+        assert events[0].http_server_request.items() >= {
+            'request_method': 'GET',
+            'path_info': '/exception',
+            'protocol': 'HTTP/1.1'
+        }.items()
+    except django.core.exceptions.PermissionDenied:
+        assert events[1].parent_id == events[0].id
+        assert events[1].exceptions[0].items() >= {
```
apotterri, line 135: To make this check look like others in the file, you should write it as:

```python
assert events[1].items() >= {
   'parent_id': events[0].id,
    'exceptions': [{
        'class': 'django.core.exceptions.PermissionDenied'
    ]}
}.items()
```

To set the `message` attribute of `exceptions[0]`, you need to pass a message to the exception constructor. For example, if you wrote:

```python
def exception_view(_request):
    raise RuntimeError('An error')
```

`exceptions[0].message` should be `'An Error'`.

apotterri (changes requested):

appmap/test/test_django.py
```diff
@@ -122,19 +121,24 @@ def test_unnamed(client, events):
     assert len(events[0].message) == 0
 
 def test_exception(client, events):
-    try:
+    with pytest.raises(RuntimeError):
         client.get('/exception')
 
+        # The 1st event: HttpServerRequestEvent
```
apotterri, line 127: These checks should be move out of the `with` block. Nothing except the request should raise a `RuntimeError`.

apotterri (changes requested):

apotterri:
Please add an entry to CHANGELOG for this fix as well.

appmap/test/test_django.py
```diff
@@ -116,6 +120,26 @@ def test_unnamed(client, events):
...
+        client.get('/exception')
+
+        # The 2nd event: ExceptionEvent
+        assert events[1].items() >= {
+            'parent_id': events[0].id,
+            'exceptions': [{
+                'class': 'RuntimeError',
+                'message': 'An error'
+            }]
+        }.items()
```
dividedmind, line 134: This could be moved out of the `with` block.

dividedmind (changes requested):
Looks good. But please use `git rebase --interactive` to squash the commits into one. No need for the commit history to reflect the development now that it's ready to merge. Also please try to use the conventional form for a commit message:

```
Change how widget operates (imperative, <60cols, no period)

This changes the widget to operate much better. This can be a 
whole proper paragraph (up to 80 cols) describing it in more detail.
You can also:
- point out various things of note here as desired,
- use bullet points and other light formatting,
- remember to make the subject succinct so it's informative,
  representative but short, '<verb> <detail>' form of what the commit does,
  suitable to understanding at a glance from `git log --graph --oneline`
  and similar commands.
```

appmap/test/test_django.py
snowmaksym: We can't check `ExceptionEvent` if the `assert` block is moved out of the `with` block.

dividedmind: I don't see why not. Have you tried it?

snowmaksym: yeah, I've tried it, but it raises `AttributeError`.
`AttributeError: 'HttpServerResponseEvent' object has no attribute 'items'`

dividedmind: Ah, right, it does. But when it's in the `with` block the code just gets skipped, because `client.get` raises.

snowmaksym: Then I think the `HttpServerResponseEvent` should also inherit the `Exception` event to check the exception from a request.

dividedmind: No. And anyway, that's beside the point. Your test just doesn't do what it's supposed to. Go ahead, remove your changes in `appmap.django`. The test still passes.

dividedmind:
@apotterri I'm not sure about these changes. I thought the whole point of #70 was to fix the issue of `get_response` raising an exception? If django catches an exception and responds with a 500, I think that should get recorded in an appmap, not an exception it has caught, even if it provides a hook for that. Also, won't using that hook potentially change the observable behavior of an app?

snowmaksym:
@apotterri, @dividedmind 
If we don't need to catch the exception from a view in the middleware, we can record the response status code 500 and its response.
Let me know I have to change the `django.py` to record the 500 response instead of an exception.

dividedmind:
> Let me know I have to change the `django.py` to record the 500 response instead of an exception.

It has already been doing this.

snowmaksym:
But I think it's better to catch the exception rather than the 500 response, because it is more helpful to understand the dynamic structure and flow of the django app.

dividedmind:
> But I think it's better to catch the exception rather than the 500 response, because it is more helpful to understand the dynamic structure and flow of the django app.

If an exception is raised in instrumented code, it will already be captured in the appmap.

apotterri:
> @apotterri I'm not sure about these changes. I thought the whole point of #70 was to fix the issue of `get_response` raising an exception? If django catches an exception and responds with a 500, I think that should get recorded in an appmap, not an exception it has caught, even if it provides a hook for that. Also, won't using that hook potentially change the observable behavior of an app?

Yeah, you're right. Let me try again.

dividedmind (approved):
Looks ok now, thanks! Please remember to fix up the commits though before merging; it seems you gave the wrong reference to `--fixup`. Also, looking at the full diff, you'll notice you don't need to qualify `django.urls.path` anymore.

appmap/test/test_django.py
```diff
@@ -150,6 +155,27 @@ def test_included_view(client, events):
...
+        'request_method': 'GET',
+        'path_info': '/exception',
+        'protocol': 'HTTP/1.1'
+    }.items()
+
+    assert events[1].parent_id == events[0].id
+    assert events[1].exceptions[0].items() >= {
+            'class': 'builtins.RuntimeError',
+            'message': 'An error'
+    }.items()
```
dividedmind, line 177: You can use DictIncluding now:
```suggestion
    assert events[0].http_server_request == DictIncluding({
        'request_method': 'GET',
        'path_info': '/exception',
        'protocol': 'HTTP/1.1'
    })

    assert events[1].parent_id == events[0].id
    assert events[1].exceptions == [DictIncluding({
            'class': 'builtins.RuntimeError',
            'message': 'An error'
    })]
```

apotterri:
> Looks ok now, thanks! Please remember to fix up the commits though before merging; it seems you gave the wrong reference to `--fixup`. Also, looking at the full diff, you'll notice you don't need to qualify `django.urls.path` anymore.

Yeah, will do. I did a rebase with `--autosquash` (because I wanted you to see a distinct fixup commit), maybe that confused things.

PR #114, opened 2021-05-26
Improve metadata

dividedmind:
- Cache git metadata
  This improves performance quite a bit since git doesn't get called several times per test case anymore.
- Streamline metadata tests.
- Skip `status` entry if it would be empty anyway.


apotterri (approved):
Looks good, thanks for doing this.

PR #116, opened 2021-05-27
Use Django cached request body when capturing JSON

dividedmind:
Fixes a problem with Django JSON parameter capture preventing the application from accessing the request body, causing the error:

    django.http.request.RawPostDataException: You cannot access body after reading from request's data stream


apotterri (approved):
Looks good, thanks.

PR #120, opened 2021-05-31
Record template rendering in Django and flask

### LINKED ISSUE #119: Record template renderings in web apps

dividedmind:
Fixes #119 .

### DIFF
--- a/appmap/flask.py
+++ b/appmap/flask.py
@@ -5,15 +5,17 @@ import time
 import flask
 import flask.cli
 from flask import _app_ctx_stack, request
+import jinja2
 from werkzeug.routing import parse_rule
 from werkzeug.exceptions import BadRequest
 
 from appmap._implementation.env import Env
 from appmap._implementation import generation
 from appmap._implementation.event import HttpServerRequestEvent, HttpServerResponseEvent
+from appmap._implementation.web_framework import TemplateHandler as BaseTemplateHandler
 from ._implementation.metadata import Metadata
 from appmap._implementation.recording import Recorder, Recording
-from ._implementation.utils import values_dict
+from ._implementation.utils import values_dict, patch_class
 
 
 try:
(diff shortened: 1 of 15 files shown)


appmap/flask.py
```diff
@@ -123,6 +125,12 @@ def after_request(self, response):
         return response
 
 
+@patch_class(jinja2.Template)
+class TemplateHandler(BaseTemplateHandler):
+    # pylint: disable=missing-class-docstring, too-few-public-methods
+    pass
```
apotterri, line 131: I thought I understood what this was doing, but seeing the `pass` here, I guess I don't.

Some of my confusion comes from having this class be named TemplateHandler, i.e. the same name that's renamed on import from `web_framework`.

I the other thing I'm confused about is whether there's a method named `render` in the `Template` class in both django and jinja2. Is there?

Can you add some comments to expand on what's happening, please?

apotterri (changes requested):
I need a little help understanding what's going on with `TemplateHandler`, but looks good otherwise.

appmap/flask.py
dividedmind: > Some of my confusion comes from having this class be named TemplateHandler, i.e. the same name that's renamed on import from `web_framework`.

I've seen this idiom in other libraries, it seems common in python. Ie. have a `lib.base.Widget` and then `lib.specific.Widget` inheriting from the former; both these names are already unambiguous and not really intended to be used together, so there's no need to make them redundant as `lib.base.BaseWidget` and `lib.specific.SpecificWidget`. Renaming `lib.base.Widget` on import to `BaseWidget` simply avoids the need to spell out `lib.base.Widget` in full in the class definition.

> I the other thing I'm confused about is whether there's a method named `render` in the `Template` class in both django and jinja2. Is there?

Yes.

> Can you add some comments to expand on what's happening, please?

Ok, I'll try :)

dividedmind: > I thought I understood what this was doing, but seeing the `pass` here, I guess I don't.

I suppose I might as well have written simply
```python
patch_class(jinja2.Template)(appmap.web_framework.TemplateHandler)
```
instead of defining a decorated empty class. Would that be any clearer?

apotterri: Ok re the rename on import. I'll get used to it, I guess. ;)

The comment you added does a good job of clarifying what's going on. With it, I think the empty class is better than the one-liner.

apotterri (approved):
Thanks for the updates.

PR #121, opened 2021-06-01
Missing config message

### LINKED ISSUE #55: Missing appmap.yml provides helpful hints
Currently, trying to generate an AppMap without an `appmap.yml` files shows a `FileNotFoundError`.

Instead, we should log a friendly message, and include some hints about what an `appmap.yml` should look like. At the very least, it should give a general example of an `appmap.yml`:

```yaml
name: app-name
packages:
- path: mod1
```

If we can offer real names (i.e. based on the current project) instead of `app-name` and `mod1`, that would be even better.

dividedmind:
Fixes #55.

Also includes a commit removing pylint exceptions. See the commit message for details.

### DIFF
--- a/appmap/_implementation/env.py
+++ b/appmap/_implementation/env.py
@@ -68,6 +68,10 @@ class Env(metaclass=_EnvMeta):
     def enabled(self):
         return self.get("APPMAP", "false").lower() == "true"
 
+    @enabled.setter
+    def enabled(self, value):
+        self.set("APPMAP", "true" if value else "false")
+
     @property
     def display_params(self):
         return self.get("APPMAP_DISPLAY_PARAMS", "true").lower() == "true"
@@ -97,7 +101,7 @@ class Env(metaclass=_EnvMeta):
                 'appmap': {
                     'level': log_level,
                     'handlers': ['default'],
-                    'propagate': False
+                    'propagate': True
                 }
             }
         }
@@ -107,7 +111,7 @@ class Env(metaclass=_EnvMeta):
                 name: {
                     'level': level.upper(),
                     'handlers': ['default'],
-                    'propagate': False
+                    'propagate': True
                 }
             })
         logging.config.dictConfig(config_dict)
--- a/pylintrc
+++ b/pylintrc
@@ -3,10 +3,10 @@
 # A comma-separated list of package or module names from where C extensions may
 # be loaded. Extensions are loading into the active Python interpreter and may
 # run arbitrary code.
-extension-pkg-whitelist=orjson
+extension-pkg-whitelist=
 
 # Specify a score threshold to be exceeded before program exits with error.
-fail-under=10.0
+fail-under=8.3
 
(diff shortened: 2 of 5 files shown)


appmap/_implementation/env.py
```diff
@@ -97,7 +101,7 @@ def _configure_logging(self):
                 'appmap': {
                     'level': log_level,
                     'handlers': ['default'],
-                    'propagate': False
+                    'propagate': True
```
apotterri, line 104: What motivated this change? IIRC, I set it to `False` because allowing the messages to propagate would cause them to be sent to the application's loggers, too. This may even have caused test failures.

pylintrc
```diff
@@ -3,10 +3,10 @@
 # A comma-separated list of package or module names from where C extensions may
 # be loaded. Extensions are loading into the active Python interpreter and may
 # run arbitrary code.
-extension-pkg-whitelist=orjson
+extension-pkg-whitelist=
 
 # Specify a score threshold to be exceeded before program exits with error.
-fail-under=10.0
+fail-under=8.3
```
apotterri, line 9: Are these changes to the pylint config actually related to fixing #55? They feel like part of a different/bigger issue. Also, it feels like we should leave the threshold set at 10.0 and fix the problems that cause failures.

apotterri (changes requested):

appmap/_implementation/env.py
dividedmind: If it doesn't propagate then it doesn't get to the pytest handler, so we can't check for the message.

I don't think it's necessarily a bad thing that appmap output gets sent to the application's loggers; unless it does break a test, of course. Perhaps we should add a flag to disable propagation?

pylintrc
dividedmind: No, I just noticed this and figured a quick improvement would be nice. See the commit message for more thoughts on this; of course it'd be better to fix the problems, however, it's potentially a lot of work and in the meantime we have those checks disabled across the board, which makes further quality degradation (under those disabled checks) invisible. So I thought it's better to reenable them and lower the threshold until we fix them. I think that's the use case this setting is intended for.

appmap/_implementation/env.py
apotterri: Ok, that makes sense. We can go with this for now, and add a flag in the future, if necessary.

pylintrc
apotterri: Ok. I guess I don't really like the whole concept of the threshold. It feels weird to me to allow the code to have a general level of meh-ness, rather than being specific about what kinds of shortcomings are acceptable.

I agree that having the checks disabled everywhere isn't good, though.

apotterri (approved):

pylintrc
dividedmind: > It feels weird to me to allow the code to have a general level of meh-ness, rather than being specific about what kinds of shortcomings are acceptable.

I agree, but allowing specific meh-ness to proliferate is IMO strictly worse ;)

PR #123, opened 2021-06-01
Fix Django path normalization, test against Django 2.2

### LINKED ISSUE #122: Path normalization in a Django app that uses `include` falis with an exception
If the `urlpatterns` for an app uses `include` to bring in a `URLconf` that contains parameters, an attempt to normalize the path of a matching request will raise an exception.

For example, the `urlpatterns` are written this way in the [chipy.org](https://github.com/chicagopython/chipy.org) app: https://github.com/chicagopython/chipy.org/blob/main/chipy_org/urls.py#L16 . This issue causes many tests to fail, including this one: https://github.com/chicagopython/chipy.org/blob/main/chipy_org/apps/job_board/tests/test_views.py#L80 .

apotterri:
Fixes #122.

### DIFF
--- a/appmap/django.py
+++ b/appmap/django.py
@@ -12,8 +12,8 @@ from django.db.backends.utils import CursorDebugWrapper
 from django.db.backends.signals import connection_created
 from django.dispatch import receiver
 from django.http import HttpResponse, HttpResponseBadRequest, JsonResponse
-from django.urls import resolve
-from django.urls.resolvers import _route_to_regex
+from django.urls import get_resolver, resolve
+
 
 import time
 import json
@@ -132,26 +132,37 @@ def request_params(request):
 
     return values_dict(params.lists())
 
+def get_resolved_match(path_info, resolver):
+    # Look through the resolver's cache of reverse
+    # route mappings and find the regex that matches the current
+    # request. The route mappings are tuples of the form:
+    #   (format string, regex, args, kwargs)
+    regex = None
+    for _,values in resolver.reverse_dict.lists():
+        for _,regex,_,_ in values:
+            match = re.match(regex, path_info[1:])
+            if match:
+                return (regex, match)
+
 def normalize_path_info(path_info, resolved):
-    # Prior to django 3.2, there doesn't appear to be a good way to
-    # determine whether the pattern that matched a request was a path
-    # or a re_path. To figure it out, just try to use the route as a
-    # regex. If that fails, assume it's a path, and use django's
-    # internal parser to turn it into a regex.
-    regex = resolved.route
-    match = re.match(regex, path_info[1:])
+    if not resolved.kwargs:
+        # No kwargs mean no parameters to normalize
+        return path_info
+
+    resolver = get_resolver()
+    regex, match = get_resolved_match(path_info, resolver)
+
+    # Given that we're processing a resolved request, there must be
+    # one that matches.
     if not match:
-        regex = _route_to_regex(resolved.route)[0]
-        match = re.match(regex, path_info[1:])
+        raise RuntimeError('No match for %s found with resolver %r' %
+                           (path_info, resolver))
 
     # Compile the regex so we can see what groups it contains.
     groups = re.compile(regex).groupindex
-    if len(groups) == 0:
-        # No named groups means no parameters to normalize.
-        return path_info
 
-    # For each parameter, insert the portion of the path that
-    # precedes, followed by the name of the parameter.
+    # For each parameter, insert the portion of the path that precedes
+    # it, followed by the name of the parameter.
     np = '/'
     pos = 0
     for i,g in enumerate(groups):
@@ -169,7 +180,10 @@ class Middleware:
         self.recorder = Recorder()
 
     def __call__(self, request):
-        if Env.current.enabled and request.path_info == '/_appmap/record':
+        if not Env.current.enabled:
+            return self.get_response(request)
+
+        if request.path_info == '/_appmap/record':
             return self.recording(request)
         if self.recorder.enabled:
             add_metadata()
@@ -240,6 +254,5 @@ def load_middleware(*args, **kwargs):
     inject_middleware()
     BaseHandler.load_middleware = original_load_middleware
     return original_load_middleware(*args, **kwargs)
-
 
 BaseHandler.load_middleware = load_middleware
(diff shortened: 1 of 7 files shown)


appmap/django.py
```diff
@@ -133,25 +133,32 @@ def request_params(request):
...
+        # No kwargs mean no parameters to normalize
+        return path_info
+
+    # Get the current resolver. Look through its cache of reverse
+    # route mappings and find the regex that matches the current
+    # request. The route mappings are tuples of the form:
+    #   (format string, regex, args, kwargs)
+    resolver = get_resolver()
+    regex = None
+    for _,regex,_,_ in resolver.reverse_dict.values():
```
dividedmind, line 146: `resolver.reverse_dict` is a multi valued dict, so `values()` will skip some items if there are views which match more than one pattern. If you use `user_view` instead of a separate `included_user_view` in the test below you can reproduce this issue.

appmap/django.py
```diff
@@ -133,25 +133,32 @@ def request_params(request):
...
-    # internal parser to turn it into a regex.
-    regex = resolved.route
-    match = re.match(regex, path_info[1:])
-    if not match:
-        regex = _route_to_regex(resolved.route)[0]
+    if not resolved.kwargs:
+        # No kwargs mean no parameters to normalize
+        return path_info
+
+    # Get the current resolver. Look through its cache of reverse
```
dividedmind, line 140: Is there a way to use `resolved.tried[-1]` (which should be the pattern that ended up matching) instead of trying to duplicate what `URLResolver.resolve` is doing?

appmap/test/test_django.py
```diff
@@ -99,23 +99,36 @@ def user_post_view(_request, username, post_id):
...
-    django.urls.path('echo', echo_view),
+    path('test', view),
+    path('', view),
+    re_path('^user/(?P<username>[^/]+)$', user_view),
+    path('post/<int:post_id>', post_view),
+    path('post/<username>/<int:post_id>/summary', user_post_view),
+    re_path(r'^post/unnamed/(\d+)$', post_unnamed_view),
+    path('echo', echo_view),
+    re_path(r'^post/(included)/', include([
+            path('<username>', user_included_view)]))
```
dividedmind, line 114: As mentioned above, if you use `user_view` here you'll catch a potential problem with the normalization code.

appmap/test/test_django.py
```diff
@@ -99,23 +99,36 @@ def user_post_view(_request, username, post_id):
...
+    assert appmap.enabled()     # sanity check
     # unnamed captures in a re_path don't show up in the event's
     # message attribute.
     assert len(events[0].message) == 0
 
+@pytest.mark.appmap_enabled
+def test_included_view(client, events):
+    client.get('/post/included/test_user')
+
+    assert len(events) > 0
```
dividedmind, line 131: Shouldn't this verify that the normalized path info ends up being `post/included/{username}`?

appmap/test/test_django.py
```diff
@@ -99,23 +99,36 @@ def user_post_view(_request, username, post_id):
...
-    django.urls.re_path(r'^post/unnamed/(\d+)$', post_unnamed_view),
-    django.urls.path('echo', echo_view),
+    path('test', view),
+    path('', view),
+    re_path('^user/(?P<username>[^/]+)$', user_view),
+    path('post/<int:post_id>', post_view),
+    path('post/<username>/<int:post_id>/summary', user_post_view),
+    re_path(r'^post/unnamed/(\d+)$', post_unnamed_view),
+    path('echo', echo_view),
+    re_path(r'^post/(included)/', include([
```
dividedmind, line 113: Why the parentheses?

dividedmind (changes requested):
I've found some corner cases where this approach can fail.

appmap/django.py
apotterri: No, because `resolved.tried` was introduced in 3.2.

appmap/test/test_django.py
apotterri: At the very least, yes. Thanks.

appmap/test/test_django.py
apotterri: I believe they're leftover from some debugging. I'll double-check. If they are actually accomplishing something, I'll add a comment to indicate what. Otherwise, I'll remove them.

appmap/django.py
dividedmind: Ah, ok, thanks. If there isn't any other way then I suppose this is acceptable, I just hope it isn't too brittle.

appmap/django.py
apotterri: Ok, thanks, I'll investigate.

appmap/django.py
apotterri: We could see if `hasattr(resolved, 'tried')` and use it if it's available, falling back to using `reverse_dict` if it's not. The 2.2 implementation seems unlikely to change at this point, and `tried` is part of the 3.2 interface. 

Does the added complexity seem worth it to you?

appmap/test/test_django.py
apotterri: Didn't need them, took them out.

appmap/test/test_django.py
apotterri: Got it, thanks.

appmap/django.py
apotterri: Well, it turns out using `tried` wasn't an option, after all. It's content for included paths isn't helpful. So, I settled on a single implementation.

pylintrc
```diff
@@ -284,7 +284,7 @@ argument-naming-style=snake_case
 
 # Regular expression matching correct argument names. Overrides argument-
 # naming-style.
-#argument-rgx=
+argument-rgx=(?P<exempt>[a-z_]{1,2})
```
dividedmind, line 287: Not sure about this. I believe there's a good reason for the rule rejecting short names in general. They're fine for very short methods (or for contexts and lambda expressions, in which cases they are already allowed), but in general the scope of a name can be long enough that a short name unnecessarily makes the code less clear. (Same is true for variable names, below).

dividedmind (approved):
Thanks, looks fine now.

But please consider reverting the short argument and variable names exemption before merging. I don't think it's a good idea.

Issue #124, opened 2021-06-04
Arguments could be captured even when signature is unknown


dividedmind:
Consider `functools.lru_cache`; this is a built-in, so signature is not available. However, we still get `args` and `kwargs` if we trace the call (as happens often when the decorator is used on a traced class). Currently in the absence of a signature parameters end up being `[]`. I think it's worthwhile to record arguments anyway.

Requires applandinc/appmap#22

kgilpin:
Would it make sense to record the arguments with arg name “args” and
“kwargs”?

Making parameter name optional because of this edge case seems aggressive.
Client tools want to be able to rely on the presence of certain fields. If
too many things are optional then client programming gets a lot harder.

dividedmind:
> Would it make sense to record the arguments with arg name “args” and “kwargs”? Making parameter name optional because of this edge case seems aggressive. Client tools want to be able to rely on the presence of certain fields. If too many things are optional then client programming gets a lot harder.

In some programming languages and environments it's simply not possible to get the parameter name. They can even possibly be unnamed. `args` and `kwargs` is just a convention. I think it'd be confusing to force it, unless it's mandated in the spec, but why work around the spec we are in control of?

kgilpin:
OK

Issue #80, opened 2021-04-14
ValueError: no signature found for builtin <functools._lru_cache_wrapper object at 0x7fffe7751768>


lovetoburnswhen:
Getting this with Python 3.7. Likely related to https://bugs.python.org/issue1748064 ?

Edit: Onlyl seems to work when I use more restrictive paths in appmap.yml

apotterri:
@Frank5000 thanks very much for reporting this. We'll need to update `appmap-python` so it expects missing signatures for built-in functions.

Can you share what project you're trying to map, and what your appmap.yml looks like?

dividedmind:
With 087e57c55418ecf961f38faaf7f6e7cd0ab93bbe lru_cache no longer raises an error. 

(Note arguments still aren't captured; with missing signature `parameters` ends up as `[]`. I've opened #124 to track this.)

PR #125, opened 2021-06-07
Follow wrappers when reflecting for function signature

dividedmind:
This fixes the problem mentioned in #124 and #80 (functions decorated with lru_cache not having their arguments captured) without actually doing #124. @apotterri, do you remember if there was a specific reason to not follow wrappers? Tests seem to pass fine, I've also ran saleor test suite and everything looks in order.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   appmap code. Instead omit the missing parameter and allow the original
   function call to raise ArgumentError if appropriate.
 - Handle the case when a method is called with self=None.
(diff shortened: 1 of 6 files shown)


apotterri:
Yeah, this was the original issue: https://github.com/applandinc/appmap-python/issues/65 . This use of `@patch` makes sure that things are working correctly: https://github.com/applandinc/appmap-python/blob/master/appmap/test/data/unittest/simple/test_simple.py#L13 . So, if the `unittest` tests are passing, it should be ok to bring back following wrappers.

Looks like the current test failures here are not related to #65.

apotterri (approved):
LGTM, thanks. Once the build is green, it's ready to go.

PR #136, opened 2021-06-19
added record decorator

Kenan7:


apotterri (changes requested):
~Hi @Kenan7. Thanks very much for this, sorry I've left it unreviewed for so long.~

~The code for the decorator looks good. Two things need to be added to the PR before it can get merged: it needs (at least) one test, and it needs a CHANGELOG entry.~

~Thanks again.~

Actually, if it's ok with you, I'll go ahead and add these. Thanks again.

Kenan7:
> Actually, if it's ok with you, I'll go ahead and add these. Thanks again.

Ok with me. Thank you.

PR #137, opened 2021-06-28
Add appmap-agent-init command

apotterri:


appmap/_implementation/configuration.py
```diff
@@ -25,19 +26,75 @@ def warn_config_missing(path):
...
+def default_app_name(rootdir):
+    rootdir = Path(rootdir)
+    if not (rootdir / '.git').exists():
+        return rootdir.name
+
+    git = utils.git(cwd=str(rootdir))
+    repo_root = git('rev-parse --show-toplevel')
+    return Path(repo_root).name
+
+def find_top_packages(rootdir):
```
torchello, line 39: I like the idea. I plan to implement something similar in Ruby as well.

appmap/_implementation/configuration.py
```diff
@@ -72,21 +129,46 @@ def name(self):
...
+            'packages': [{'path': p} for p in find_top_packages(root_dir)],
+        }
+
     @property
     @lru_cache(maxsize=None)
     def _config(self):
-        path = Path(Env.current.get("APPMAP_CONFIG", "appmap.yml")).resolve()
+        # Only use a default config if the user hasn't specified a
+        # config.
+        env_config = Env.current.get("APPMAP_CONFIG")
```
torchello, line 146: Do we need to support `APPMAP_CONFIG` env var for code extensions? Feels like the initial flow would not ask user to set env vars.

torchello: Ruby CLI supports passing `-c,` `--config` params to set custom path though. I think we could focus on using the default path in the first phase.

torchello (approved):
Left comment regarding the usage of env vars.

kgilpin:
APPMAP_CONFIG is not required

appmap/_implementation/configuration.py
apotterri: `APPMAP_CONFIG` only needs to be used at runtime, the CLI doesn't need to support it (as Kevin said).

We can add support for `--config` to `appmap-agent-init` when the code extensions need it.

PR #139, opened 2021-07-06
Make django path normalization more robust

### LINKED ISSUE #138: Path normalization fails in misago
The routing in https://github.com/rafalp/Misago confuses the current path normalization, causing it to fail:

```
path_info = '/a/secret/1/'
resolved = ResolverMatch(func=misago.threads.views.attachment.attachment_server, args=(), kwargs={'secret': 'secret', 'pk': '1'}, url_name=attachment, app_names=['misago'], namespaces=['misago'], route=^a/(?P<secret>[-a-zA-Z0-9]+)/(?P<pk>\d+)/)

    def normalize_path_info(path_info, resolved):
        if not resolved.kwargs:
            # No kwargs mean no parameters to normalize
            return path_info

        resolver = get_resolver()
>       regex, match = get_resolved_match(path_info, resolver)
E       TypeError: cannot unpack non-iterable NoneType object

/src/appmap-python/appmap/django.py:160: TypeError
```

apotterri:
Make path normalization more robust. Instead of trying to duplicate the
functionality of resolve(), we'll call it and interpret the route it
returns.

Fixes #138.

### DIFF
--- a/appmap/django.py
+++ b/appmap/django.py
@@ -14,6 +14,8 @@ from django.dispatch import receiver
 from django.http import HttpResponse, HttpResponseBadRequest, JsonResponse
 from django.template import Template
 from django.urls import get_resolver, resolve
+from django.urls.resolvers import _route_to_regex
+from django.urls.exceptions import Resolver404
 
 import time
 import json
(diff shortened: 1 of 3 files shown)


appmap/django.py
```diff
@@ -140,16 +142,27 @@ def request_params(request):
...
-    # request. The route mappings are tuples of the form:
-    #   (format string, regex, args, kwargs)
-    regex = None
-    for _,values in resolver.reverse_dict.lists():
-        for _,regex,_,_ in values:
-            match = re.match(regex, path_info[1:])
-            if match:
-                return (regex, match)
+    resolver_match = resolver.resolve(path_info)
+    regex = resolver_match.route
```
dividedmind, line 146: It's a bit confusing that this variable is called `regex`. Maybe rename to `route`?

appmap/django.py
```diff
@@ -140,16 +142,27 @@ def request_params(request):
...
+
+    # Sometimes the route returned by resolve is a regex, sometimes
+    # it's a url pattern. If it's a regex, it will either have a named
+    # group, or it will end with a '$'. If it's a url pattern, convert
```
dividedmind, line 150: Is that a fact? I'm wary that it'd turn out to be another assumption that isn't always true. Is it possible to look at the type, does it return a typed regex or a textual representation?

dividedmind (approved):
I've been always a bit wary of that reimplementation. I hoped it'll work though ;)

Couple comments, but definitely an improvement.

appmap/django.py
apotterri: I guess. The unfortunate thing is that `resolver_match.route` *is* a regex sometimes, and the ultimate goal here is to make sure we have a regex that matches `path_info`.

appmap/django.py
apotterri: TBH, I'm not 100% sure it's a fact. I'm going to walk through the code, as well as add some tests, to try to figure it out.

`resolver_match.route` is a string. AFAIK, Python doesn't have a regex type.

torchello (approved):

PR #142, opened 2021-07-09
Make sure the appmap django middleware is always present

### LINKED ISSUE #141: Only one http_server_request captured for saleor
When mapping the tests in [saleor](https://github.com/mirumee/saleor), only the first HTTP request generates an `http_server_request` event. As it happens, the first test that makes an HTTP request is this one https://github.com/mirumee/saleor/blob/master/saleor/core/tests/test_middleware.py, which removes `appmap.django.Middleware` from`django.conf.settings.MIDDLEWARE`.

apotterri:
Make sure that our django middleware is always present in the stack, even if a test resets it.

Also, factor the django test app out from test_django.py into a separate app.

Fixes #141.

### DIFF
--- a/appmap/django.py
+++ b/appmap/django.py
@@ -288,7 +288,6 @@ original_load_middleware = BaseHandler.load_middleware
 def load_middleware(*args, **kwargs):
     """Patched wrapper to inject AppMap middleware first"""
     inject_middleware()
-    BaseHandler.load_middleware = original_load_middleware
     return original_load_middleware(*args, **kwargs)
 
 BaseHandler.load_middleware = load_middleware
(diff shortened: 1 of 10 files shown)


appmap/django.py
```diff
@@ -288,7 +288,6 @@ def inject_middleware():
 def load_middleware(*args, **kwargs):
     """Patched wrapper to inject AppMap middleware first"""
     inject_middleware()
-    BaseHandler.load_middleware = original_load_middleware
```
apotterri, line 291: @divide resetting to the original `load_middleware` function here causes problems if the test sets the middleware (as happens here https://github.com/mirumee/saleor/blob/master/saleor/core/tests/test_middleware.py#L16).

Did you do this just for the sake of efficiency, or was there some other reason for it?

torchello (approved):
Looks ok!

kgilpin:
Is this done?

apotterri:
Yup.

PR #143, opened 2021-07-13
Add appmap-agent-status command

apotterri:
The subject says it all. :)

Note that discovery of `unittest` tests is not yet supported.

Fixes #143 .


torchello (approved):
Looks good! I'd split `appmap/command/appmap_agent_status.py` into multiple classes to allow reusing the logic in the `appmap-agent-validate` command, but could be done when implementing it.

apotterri:
Thanks. It doesn't look to me like they have that much in common. I'll certainly refactor and share whatever code I can when it comes time to implement `appmap-agent-validate`.

PR #144, opened 2021-07-22
Move test discovery to appmap-agent-init

apotterri:
Move test discovery from `appmap-agent-status` to `appmap-agent-init`.

### DIFF
--- a/appmap/test/data/agent_status/appmap.yml
+++ b/appmap/test/data/agent_status/appmap.yml
@@ -1,0 +1,9 @@
+name: agent_status_test
+packages: []
+test_commands:
+- framework: "pytest"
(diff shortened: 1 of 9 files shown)


appmap/test/data/agent_status/appmap.yml
```diff
@@ -0,0 +1,9 @@
+name: agent_status_test
+packages: []
+test_commands:
```
torchello, line 3: Is this just a small hack for Python or we going to update the docs and other language agents as well? /cc @kgilpin

kgilpin:
I think we should hold off on this for now since we are reworking
other parts of the interface.

PR #148, opened 2021-09-27
fix: Fix validation for web frameworks

apotterri:
Both Django and Flask have two supported major versions. Fix validation
and update tests to reflect this.

### DIFF
--- a/appmap/__init__.py
+++ b/appmap/__init__.py
@@ -10,6 +10,11 @@ except ImportError:
     # not using flask
     pass
 
+try:
+    from . import django
+except ImportError:
+    # not using django
+    pass
 
 def enabled():
     return Env.current.enabled
--- a/appmap/command/appmap_agent_validate.py
+++ b/appmap/command/appmap_agent_validate.py
@@ -23,31 +23,37 @@ def check_python_version():
     except AppMapPyVerException as e:
         raise ValidationFailure(str(e))
 
-def _check_version(dist, v):
+def _check_version(dist, versions):
     dist_version = None
-    try:
-        dist_version = version(dist)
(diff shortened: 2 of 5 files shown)


appmap/command/appmap_agent_validate.py
```diff
@@ -23,31 +23,37 @@ def check_python_version():
...
+            if actual < required:
+                raise ValidationFailure(f'{dist} must have version >= {required}, found {actual}')
 
-        if actual < required:
-            raise ValidationFailure(f'{dist} must have version >= {required}, found {actual}')
-    except PackageNotFoundError:
-        pass
+            return dist_version
+        except PackageNotFoundError:
+            dist_version = None
```
dividedmind, line 43: Nitpick: this changed logic is a bit convoluted and difficult to follow. How about something like:
```python
try:
  dist_version = version(dist)
  actual = parse(dist_version)
  required = next(v for v in map(parse, versions) if v.major == actual.major)
  if actual < required:
    raise ValidationFailure
except PackageNotFoundError, StopIteration:
  return None
return dist_version
```

appmap/__init__.py
```diff
@@ -10,6 +10,11 @@
     # not using flask
     pass
 
+try:
+    from . import django
+except ImportError:
+    # not using django
+    pass
```
dividedmind, line 17: Seems ok to me, python isn't as fragile re import order as ruby can be.

dividedmind (commented):
Looks good, but note this changes the semantics and it's not clear to me this is intended. In particular, consider what happens when a dependency is detected with a major version different from any on the list of supported ones.

PR #153, opened 2021-12-06
Allow specifying function labels in the config file

dividedmind:
This change is especially useful for library functions; it's not intended to replace the @labels decorator but instead allow to configure labeling of code outside of the developer control.

Planned followups:
- doc change on the website,
- predefined labels for popular libraries.

### DIFF
--- a/appmap/_implementation/recording.py
+++ b/appmap/_implementation/recording.py
@@ -90,7 +90,7 @@ class FilterableFn(
     __slots__ = ()
 
     def __new__(c, scope, fn, static_fn):
-        fqname = '%s.%s' % (scope.fqname, fn.__qualname__)
+        fqname = '%s.%s' % (scope.fqname, fn.__name__)
         self = super(FilterableFn, c).__new__(c, fqname, fn, scope, static_fn)
         return self
 
--- a/appmap/_implementation/configuration.py
+++ b/appmap/_implementation/configuration.py
@@ -2,6 +2,7 @@
 Manage Configuration AppMap recorder for Python.
 """
 
+from functools import lru_cache
 import inspect
 import logging
 from os.path import realpath
@@ -17,6 +18,7 @@ from yaml.parser import ParserError
 from . import utils
 from .env import Env
 from .instrument import instrument
+from .labels import labels
 from .metadata import Metadata
 from .recording import Filter, FilterableCls, Recorder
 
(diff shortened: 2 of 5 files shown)


kgilpin:
Looks good. @ptrdvrk this will be relevant to you as well.

appmap/_implementation/recording.py
```diff
@@ -90,7 +90,7 @@ class FilterableFn(
     __slots__ = ()
 
     def __new__(c, scope, fn, static_fn):
-        fqname = '%s.%s' % (scope.fqname, fn.__qualname__)
+        fqname = '%s.%s' % (scope.fqname, fn.__name__)
```
apotterri, line 93: Did you see a failure of some kind that motivated this change? If not, I don't think it's right. IIRC, the strings in `scope.fqname` are all module names, but we also need the name of the class that contains the function. `__qualname__` gives us the class, `__name__` doesn't:
```
ajp@Alans-MacBook-Pro appmap-python % python
Python 3.9.7 (default, Sep 27 2021, 17:48:06)
[Clang 12.0.5 (clang-1205.0.22.11)] on darwin
>>> import appmap._implementation.configuration;
>>> appmap._implementation.configuration.PathMatcher.matches.__name__
'matches'
>>> appmap._implementation.configuration.PathMatcher.matches.__qualname__
'PathMatcher.matches'
>>>
```

No matter what, seems like we're missing a test, since this change didn't cause any failures....

apotterri (changes requested):
The functionality related to labels looks good. I'm not sure we want the change to `FilterableFn`.

appmap/_implementation/recording.py
dividedmind: Yes, but the `scope` already is the class scope. I noticed it while adding matching of functions for the labeling. `fqname`s ended up being, for example, `example_class.ExampleClass.ExampleClass.test_exception`.
Re no failures, that led me to assume `fqname`s weren't used for anything (anymore?); addition of labels to the integration test now exercises them, so without this change they would fail.

apotterri: Hmmm, ok, `example_class.ExampleClass.ExampleClass.test_exception` is clearly not right.

I think we are still using `fqname`, here: https://github.com/applandinc/appmap-python/blob/master/appmap/_implementation/configuration.py#L226 . But, as far as I can tell, we have tests that cover that code. I guess if they still pass with the change, it's probably ok.

apotterri (approved):

PR #154, opened 2021-12-16
Preset labels for known library functions

dividedmind:
Just the infrastructure. Specific labels for more libraries to come.

### DIFF
--- a/appmap/_implementation/configuration.py
+++ b/appmap/_implementation/configuration.py
@@ -18,7 +18,7 @@ from yaml.parser import ParserError
 from . import utils
 from .env import Env
 from .instrument import instrument
-from .labels import labels
+from .labels import LabelSet
 from .metadata import Metadata
 from .recording import Filter, FilterableCls, Recorder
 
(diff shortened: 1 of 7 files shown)


dividedmind:
One thing I couldn't decide: do you think the presence of a label declaration should automatically cause the function to be recorded? appmap-ruby does this (I believe). In appmap-python currently labeling and recording are separate settings -- the functions still need to be added to the config file `packages` field.

apotterri (approved):
Looks good, thanks.

apotterri:
> One thing I couldn't decide: do you think the presence of a label declaration should automatically cause the function to be recorded? appmap-ruby does this (I believe). In appmap-python currently labeling and recording are separate settings -- the functions still need to be added to the config file `packages` field.

I don't know what appmap-ruby's behavior is. Following whatever it does seems fine to me.

kgilpin:
`appmap-ruby` doesn't automatically record anything that's labeled. It would have to parse every source file of every Gem and that seems infeasible.

dividedmind:
> `appmap-ruby` doesn't automatically record anything that's labeled. It would have to parse every source file of every Gem and that seems infeasible.

But doesn't it always hook and record whatever has labels defined on in `hooks`?

kgilpin:
Yes, it does do that. I thought you meant it would record any source file
with @label

PR #156, opened 2022-01-28
fix: Handle a partial finder function

apotterri:
If a finder function doesn't have a `__name__` attr, and it's a
`functools.partial`, use its `func.__name__` instead.

### DIFF
--- a/appmap/test/data/unittest/simple/test_simple.py
+++ b/appmap/test/data/unittest/simple/test_simple.py
@@ -1,6 +1,10 @@
 import unittest
 from unittest.mock import patch
 
+# Importing from decouple will cause a failure if we're not hooking
+# finders correctly.
+from decouple import config
+
 import appmap.unittest
 
 import simple
--- a/appmap/unittest.py
+++ b/appmap/unittest.py
@@ -1,43 +1,56 @@
 from contextlib import contextmanager
+import logging
 import unittest
 
+import appmap
 from appmap._implementation import testing_framework
 import appmap.wrapt as wrapt
 
 
-session = testing_framework.session('unittest')
+logger = logging.getLogger(__name__)
 
-def get_test_location(cls, method_name):
-    from appmap._implementation.utils import get_function_location
-    fn = getattr(cls, method_name)
-    return get_function_location(fn)
-
-
-# unittest.case._Outcome.testPartExecutor is used by all supported
-# versions of unittest to run test cases. `isTest` will be True when
-# the part is the actual test method, False when it's setUp or
-# teardown.
-@wrapt.patch_function_wrapper('unittest.case', '_Outcome.testPartExecutor')
-@contextmanager
-def testPartExecutor(wrapped, _, args, kwargs):
-    def _args(test_case, *_, isTest=False, **__):
-        return (test_case, isTest)
-    test_case, is_test = _args(*args, **kwargs)
-    already_recording = getattr(test_case, '_appmap_pytest_recording', None)
-    if (not is_test) or already_recording:
-        with wrapped(*args, **kwargs):
-            yield
+def setup_unittest():
+    if not appmap.enabled():
+        # Using this runner without enabling AppMap might not be what
+        # the user intended, so issue a warning.
+        logger.warning("AppMap disabled. Did you forget to set APPMAP=true?")
         return
 
-    method_name = test_case.id().split('.')[-1]
-    location = get_test_location(test_case.__class__, method_name)
-    with session.record(
-            test_case.__class__,
-            method_name,
-            location=location) as metadata:
-        with wrapped(*args, **kwargs), testing_framework.collect_result_metadata(metadata):
-            yield
+    session = testing_framework.session('unittest')
 
+    def get_test_location(cls, method_name):
+        from appmap._implementation.utils import get_function_location
+        fn = getattr(cls, method_name)
+        return get_function_location(fn)
+
+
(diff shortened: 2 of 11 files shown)


dividedmind:
This seems sensible, but what's missing from the commit message is an example of when a finder would be a wrapped function.

apotterri:
Yeah, I realized that right after I tagged you to review, which is why I switched it to Draft.

apotterri:
> This seems sensible, but what's missing from the commit message is an example of when a finder would be a wrapped function.

Turns out that `functools.partial` doesn't set `__wrapped__`, so `inspect.unwrap` doesn't returned the wrapped function. So, instead, I resorted to checking if the function is `isinstance(fn,functools.partial)`.

apotterri:
Here are these changes in (almost) installable form:
[appmap-1.5.1.dev1-py3-none-any.whl.zip](https://github.com/applandinc/appmap-python/files/7974303/appmap-1.5.1.dev1-py3-none-any.whl.zip)
Unzip to get an actual wheel. Thanks GitHub. :|

dividedmind:
This looks better now. However, reading the code more closely I don't understand why the marker is even necessary -- the wrapped finder is directly returned, so won't this return an unwrapped finder next time the same one gets inserted? Or in case of a partial function -- next time a similar finder is used? Surely we want to wrap all finders in the meta path, especially if they're different partial applications of the same function. The issue of unwrapped finders is avoided by the fact that the markers are set directly on the `Partial` objects themselves. But again, why the marker is even necessary?

appmap/test/data/unittest/simple/test_simple.py
```diff
@@ -1,6 +1,10 @@
 import unittest
 from unittest.mock import patch
 
+# Importing from decouple will cause a failure if we're not hooking
+# finders correctly.
+from decouple import config
```
dividedmind, line 6: What makes `decouple` special here?

ci/test_install.sh
```diff
@@ -7,3 +7,10 @@ docker run -i --rm\
   --entrypoint /bin/bash\
   -w /tmp\
   python:3.9 /ci/smoketest.sh
+
+docker run -i --rm\
+  -v $PWD/dist:/dist -v $PWD/appmap/test/data/unittest:/appmap/test/data/unittest\
+  -v $PWD/ci:/ci\
+  --entrypoint /bin/bash\
+  -w /tmp\
+  python:3.9 /ci/test_pipenv.sh
```
dividedmind, line 16: Perhaps these scripts could be ran in a single `docker run` instance.

appmap/unittest.py
```diff
@@ -1,43 +1,56 @@
...
-            test_case.__class__,
-            method_name,
-            location=location) as metadata:
-        with wrapped(*args, **kwargs), testing_framework.collect_result_metadata(metadata):
-            yield
+def setup_unittest():
+    if not appmap.enabled():
+        # Using this runner without enabling AppMap might not be what
+        # the user intended, so issue a warning.
+        logger.warning("AppMap disabled")
```
dividedmind, line 16: I think the message should tell the user how to enable it.

appmap/unittest.py
```diff
@@ -1,43 +1,56 @@
...
 
-    method_name = test_case.id().split('.')[-1]
-    location = get_test_location(test_case.__class__, method_name)
-    with session.record(
-            test_case.__class__,
-            method_name,
-            location=location) as metadata:
-        with wrapped(*args, **kwargs), testing_framework.collect_result_metadata(metadata):
-            yield
+def setup_unittest():
```
dividedmind, line 12: What's the purpose of wrapping this in a function?

dividedmind (commented):
This looks better, but reading the code I'm not sure I understand why the marker is necessary or even useful. 
- We're returning the decorated function directly -- how could double wrapping happen? 
- Why not just mark the decorated function itself (eg. with a static `._appmapped_finder=True` or by making it a callable object of a specific class)?
- Won't the marker leave some finders unwrapped if they're duplicated on the meta path? 
- And why does the marker need the function name at all -- especially when it's not a bound method (in which case the marker is (for some reason) placed on the object, which could have presumably several finder methods)?

I tried looking at past commits to help me understand this, but it didn't help much. It seems the code could be significantly simplified, or at least use some comments.

appmap/unittest.py
apotterri: To make it easy to avoid doing the setup if `not appmap.enabled()`.

appmap/test/data/unittest/simple/test_simple.py
apotterri: I'm not sure. It was in the stack trace provided by the user that ran into the problem. Its presence here reproduced the failure before I applied the fix

appmap/unittest.py
apotterri: Ok. I'll add some more text.

ci/test_install.sh
apotterri: Ok, it should be straightforward to combine them.

apotterri:
I initially tried removing the function name from the marker. That caused several tests to fail, so I went with this change.

TBH, I have other work that's higher-priority than cleaning up this code. My fix doesn't change the way it functions beyond how the function name is obtained. So, in all other respects, it's as good as it was before.

I'll make the two simple changes I indicated, but I'd like to leave the rest of the code as-is.

kgilpin (approved):

PR #159, opened 2022-06-22
fix: Avoid querying database version when executing client queries

### LINKED ISSUE #158: `Cannot recover from stack overflow` error running pytest with django and mysql
as the title described.  Thanks

dividedmind:
While I couldn't reproduce it, a user has reported seeing a stack
overflow while processing an SQL query with MySQL and django. It
seems to have been caused by appmap querying for the server version,
which entails another SQL query against the server with MySQL.

Since MySQL backend caches that information, querying it when first
establishing the connection ensures the additional query is avoided
later at runtime, thus avoiding this problem.

Fixes #158


apotterri (approved):
LGTM. Thanks!

PR #171, opened 2022-09-26
Per thread recording

apotterri:
When creating a Recorder, allow the caller to specify a thread id. When
new events are added, they will be separated based on the thread id in
the event itself.

The `refactor` touches a lot of files, but it's just the result of running `black` and `isort` on the source. It probably doesn't need much (or any) review.

Fixes getappmap/board#169.

### DIFF
--- a/appmap/_implementation/recording.py
+++ b/appmap/_implementation/recording.py
@@ -6,13 +6,14 @@ import functools
 import inspect
 import logging
 import sys
+import threading
 import traceback
 import types
 
 import appmap.wrapt as wrapt
 
 from .env import Env
-from .event import _EventIds
+from .event import _EventIds, Event
 from .metadata import Metadata
 from .utils import FnType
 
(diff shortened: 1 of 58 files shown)


appmap/_implementation/recording.py
```diff
@@ -188,35 +189,46 @@ def is_member_func(m):
 
 
 class Recorder:
-    """ Singleton Recorder class """
-    _instance = None
+
```
symwell: Note
- in `appmap-ruby` Tracing contains many Tracers, one per thread.
- in `appmap-python` Recording contains one Recorder. The Recorder saves for many threads.

The two designs don't match.  I'm not suggesting changing the `appmap-python` design.

Is it ok for the two designs to not match?

apotterri: I don't think the designs need to match. Both seem clear to me, and the approach I took required the fewer changes to the source.

symwell (approved):

appmap/_implementation/recording.py
apotterri: I don't think the designs need to match. Both seem clear to me, and the approach I took required the fewer changes to the source.

PR #175, opened 2022-10-01
Fix problems mapping with python 3.10

apotterri:
Fixes getappmap/board#150.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
-language: python
-python: 3.9
+dist: jammy
+addons:
+  apt:
+    update: true
+    packages: [build-essential zlib1g-dev libncurses5-dev libgdbm-dev libnss3-dev libssl-dev libreadline-dev libffi-dev libsqlite3-dev wget libbz2-dev]
 
 # https://github.com/travis-ci/travis-ci/issues/1147#issuecomment-441393807
 if: type != push OR branch = master OR branch =~ /^v\d+\.\d+(\.\d+)?(-\S*)?$/
@@ -11,17 +14,20 @@ env:
 - PYENV_ROOT="$HOME/.pyenv"
 
 install: |
+  set -e
   if [[ ! -d "$PYENV_ROOT/versions" ]]; then
     rmdir "$PYENV_ROOT"
     curl -L https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash
+    git clone https://github.com/momo-lab/xxenv-latest.git "$(pyenv root)/plugins/xxenv-latest"
   fi
   export PATH="$PYENV_ROOT/bin:$PATH"
+  eval "$(pyenv init -)"
+  eval "$(pyenv init --path)"
+  eval "$(pyenv virtualenv-init -)"
   pyenv update
-  git clone https://github.com/momo-lab/xxenv-latest.git "$(pyenv root)/plugins/xxenv-latest"
-  for v in {7..9}; do pyenv latest install -s 3.$v; pyenv latest uninstall -f 3.$v; done
-  pyenv latest local 3.{7..9}
-  pip install --upgrade pip
-  pip install poetry tox
+  for v in {7..10}; do pyenv latest install -s 3.$v; pyenv latest uninstall -f 3.$v; done
+  pyenv latest local 3.{7..10}
+  pip install --upgrade pip poetry tox
 
 
 cache:
--- a/appmap/_implementation/recording.py
+++ b/appmap/_implementation/recording.py
@@ -286,6 +286,7 @@ class Recorder:
         if mod.__name__.startswith("appmap"):
             return
 
+        print("mod %s", mod)
         logger.debug("do_import, mod %s args %s kwargs %s", mod, args, kwargs)
         if not cls.filter_chain:
             logger.debug("  filter_stack %s", cls.filter_stack)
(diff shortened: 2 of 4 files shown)


appmap/_implementation/recording.py
```diff
@@ -355,7 +355,18 @@ def wrapped_find_spec(find_spec, _, args, kwargs):
     if spec is not None:
         if getattr(spec.loader, "exec_module", None) is not None:
             loader = spec.loader
-            loader.exec_module = wrap_exec_module(loader.exec_module)
+            # This is kind of gross. As this comment linked to below describes, wrapt has trouble
+            # identifying methods decorated with @staticmethod. It offers two suggested fixes:
+            # update the class definition, or patch the function in __dict__. We can't do the
+            # former, so do the latter instead.
+            #   comment in wrapt: https://github.com/GrahamDumpleton/wrapt/blob/master/src/wrapt/wrappers.py#L730
```
dividedmind, line 362: This should probably use a permalink with commit id. You can easily get one by pressing the 'Y' key in github.

dividedmind (approved):
Looks good. Had one comment (on a comment, on a comment).

BTW, I wonder if it wouldn't be a good idea to take a good look at the descriptor protocol and see if we can roll our own code to improve on what wrapt is doing for us. It sometimes seems it has almost as many corner cases we have to handle as we'd have to when doing it directly.

.travis.yml
```diff
@@ -11,17 +14,20 @@ env:
...
+  eval "$(pyenv init -)"
+  eval "$(pyenv init --path)"
+  eval "$(pyenv virtualenv-init -)"
   pyenv update
-  git clone https://github.com/momo-lab/xxenv-latest.git "$(pyenv root)/plugins/xxenv-latest"
-  for v in {7..9}; do pyenv latest install -s 3.$v; pyenv latest uninstall -f 3.$v; done
-  pyenv latest local 3.{7..9}
-  pip install --upgrade pip
-  pip install poetry tox
+  for v in {7..10}; do pyenv latest install -s 3.$v; pyenv latest uninstall -f 3.$v; done
```
dividedmind, line 28: I wonder, is there a distro or another way in travis to have all these pythons precompiled? If not, maybe we could pull binaries? This really seems a huge waste of cycles.

appmap/_implementation/recording.py
```diff
@@ -355,7 +355,18 @@ def wrapped_find_spec(find_spec, _, args, kwargs):
...
+            #
+            # TODO: determine if we can use wrapt.wrap_function_wrapper to simplify this code
+            exec_module = inspect.getattr_static(loader, "exec_module")
+            if isinstance(exec_module, staticmethod):
+                loader.exec_module = wrap_exec_module(exec_module)
```
dividedmind, line 367: I don't entirely understand how it's different. Surely if `exec_module` is a `staticmethod`, `loader.exec_module === exec_module`? Or is there something else going on?

dividedmind: Ah, never mind, actually `loader.exec_module == exec_module.__wrapped__` (or whatever is the attr called).

appmap/_implementation/recording.py
apotterri: Yes, you're right. GH even did the right thing when I copied the link, and I changed it `master`. Oops. I'll change it back.

.travis.yml
apotterri: The config is supposed to be telling Travis to cache the `pyenv` installs:
```
cache:
- $PYENV_ROOT
```
and this
```
for v in {7..10}; do pyenv latest install -s 3.$v; pyenv latest uninstall -f 3.$v; done
```
is supposed to be ensuring that just the latest release is in the cache: https://github.com/momo-lab/xxenv-latest#usage .

After I get this PR merged, I'll investigate why this isn't working, and improve it (somehow).

appmap/_implementation/recording.py
apotterri: Yeah:
```
ajp@Alans-MacBook-Pro appmap-python % python
Python 3.10.7 (main, Sep 15 2022, 04:53:36) [Clang 13.1.6 (clang-1316.0.21.2.5)] on darwin
>>> from importlib.machinery import BuiltinImporter
>>> import inspect
>>> getattr(BuiltinImporter, 'exec_module')
<function BuiltinImporter.exec_module at 0x10edde4d0>
>>> inspect.getattr_static(BuiltinImporter, 'exec_module')
<staticmethod(<function BuiltinImporter.exec_module at 0x10edde4d0>)>
```

.travis.yml
apotterri: FWIW, builds on the same branch do cache $PYENV_ROOT correctly:
https://app.travis-ci.com/github/getappmap/appmap-python/jobs/584499355#L262 .

The build for the merged PR didn't use the cache though. :( So, there's definitely room for improvement here.

apotterri:
> Looks good. Had one comment (on a comment, on a comment).
> 
> BTW, I wonder if it wouldn't be a good idea to take a good look at the descriptor protocol and see if we can roll our own code to improve on what wrapt is doing for us. It sometimes seems it has almost as many corner cases we have to handle as we'd have to when doing it directly.

Yeah, I agree that we might be better off without `wrapt`. Coming up with a different strategy is on my list of things to do post Tech Crunch.

.travis.yml
dividedmind: Thanks for looking into this!

Issue #166, opened 2022-09-19
get python agent to record by default


symwell:
Similar pr in `appmap-ruby` https://github.com/getappmap/appmap-ruby/pull/281

Looked at Django and Flask, there are notes
Starting here - https://applanding.slack.com/archives/CQ5JFE1DY/p1663189579942259

About dev / test / prod in Django - https://stackoverflow.com/questions/10664244/django-how-to-manage-development-and-production-settings - tldr it's not really a standardized env var like rails.

Flask has ENV, DEBUG, TESTING and some other stuff - https://flask.palletsprojects.com/en/2.2.x/config/
app.config[“APPMAP”] might be the best option
Configuration is expected to be these flags on the app and then the code acts accordingly
So in other words it may be more idiomatic to check app.config rather than the environment
It's up to the developers to set those flags using whatever scheme they choose

apotterri:
Discussion about this here: https://applanding.slack.com/archives/CQ5JFE1DY/p1664381082288939

```
Going forward, I think we would like to support two ways of enabling AppMap for Django (
@Alan
 correct me if I’m wrong) (edited) 
1) APPMAP=true - enables all the recordings
2) Using Django “settings” - if the user loads a settings file that contains something like AppMap = True then AppMap will be enabled, unless APPMAP=false. (edited) 
This would be pretty consistent with Rails, where request + remote recording is auto-enabled in development, unless APPMAP=false, and test cases recording is auto-enabled in test, unless APPMAP=false
An individual env var is also available to enable/disable each recording method, like APPMAP_RECORD_RSPEC=(true|false)
This is not really for normal usage though
It’s all documented on the AppMap doc site for appmap-ruby, so check that and see if anything is missing and/or unclear
The main point is, using settings files seems to be idiomatic for Python, so we can support that. According to Alan, using env vars is also fairly common, so we support that too.
Ideally, we’d like developers to put AppMap = True in their development and test settings so it’s just on all the time. (edited)
```

PR #176, opened 2022-10-04
feat: Record by default

symwell:
Record appmaps by default.

Fixes https://github.com/getappmap/appmap-python/issues/166

### DIFF
--- a/appmap/flask.py
+++ b/appmap/flask.py
@@ -78,9 +78,8 @@ class AppmapFlask(AppmapMiddleware):
                 methods=["DELETE"],
             )
 
-        if DetectEnabled.should_enable("requests"):
-            app.before_request(self.before_request)
-            app.after_request(self.after_request)
+        app.before_request(self.before_request)
+        app.after_request(self.after_request)
 
     def record_get(self):
         if not self.should_record():


appmap/test/test_detect_enabled.py
```diff
@@ -0,0 +1,143 @@
...
+# AttributeError: 'Settings' object has no attribute 'ROOT_URLCONF'
+ROOT_URLCONF = "app.urls"
+
+MIDDLEWARE = ["appmap.django.Middleware"]
+
+"""
+
+    @patch.dict(os.environ, {"DJANGO_SETTINGS_MODULE": "app.settings_debug_false"})
+    def test_django__debug_false(self, data_dir, monkeypatch):
+        os.environ.setdefault("PYTHONPATH", os.getcwd())
```
apotterri, line 125: `monkeypatch` provides a way to make changes to the environment, which is used by other tests.
It would better to use that, rather than using either `@patch`,  or making any changes to `os.environ` directly.

So, this code can be written as
```
     def test_django__django_debug_false(self, data_dir, monkeypatch):
       monkeypatch.syspath_prepend(data_dir / "django")
       monkeypatch.setenv("DJANGO_SETTINGS_MODULE", "app.settings_debug_false")
```

appmap/test/test_detect_enabled.py
```diff
@@ -0,0 +1,143 @@
...
+        filename = data_dir / "django" / "app" / "settings_debug_false.py"
+        file_write(str(filename), TestDetectEnabled.settings_most() + """DEBUG = False
+""")
+        monkeypatch.chdir(str(data_dir / "django"))
+        DetectEnabled.initialize(None)
+        assert DetectEnabled.should_enable() == False
+        file_delete(filename)
```
apotterri, line 132: This will leave behind the file if the `assert` fails, so you could do this to ensure it gets cleaned up.

```suggestion
        try:
            monkeypatch.chdir(str(data_dir / "django"))
            DetectEnabled.initialize(None)
            assert DetectEnabled.should_enable() == False
        finally:
            file_delete(filename)
```

appmap/_implementation/detect_enabled.py
```diff
@@ -0,0 +1,135 @@
...
+                settings = importlib.import_module(os.environ['DJANGO_SETTINGS_MODULE'])
+            except Exception as exn:
+                settings = None
+                return ["couldn't load DJANGO_SETTINGS_MODULE={os.environ['DJANGO_SETTINGS_MODULE']}", False]
+
+            if settings:
+                try:
+                    if settings.DEBUG == True:
+                        return [f"{os.environ['DJANGO_SETTINGS_MODULE']}.DEBUG={os.environ['DEBUG']}", True]
+                except:
```
apotterri, line 123: Better not to write bare `except`s. This code raises an exception when accessing `os.environ['DEBUG']`, and causes the function to incorrectly return `["it is not Django", False]`.

appmap/_implementation/testing_framework.py
```diff
@@ -147,3 +147,14 @@ def collect_result_metadata(metadata):
...
+
+def file_write(filename, contents):
+    f = open(filename, "w")
+    f.write(contents)
+    f.close()
+
+def file_delete(filename):
+    try:
+        os.remove(filename)
+    except:
```
apotterri, line 159: This bare except hides the error that `os` hasn't been imported. It should just be catching `FileNotFoundError`, and allowing any other exception to propagate.

appmap/test/test_detect_enabled.py
```diff
@@ -0,0 +1,143 @@
...
+from appmap._implementation.detect_enabled import DetectEnabled, RECORDING_METHODS
+from appmap._implementation.testing_framework import file_write, file_delete
+
+class TestDetectEnabled:
+    @staticmethod
+    def setup_method():
+        DetectEnabled.clear_cache()
+
+    @patch.dict(os.environ, {"APPMAP": "false"})
+    def test_none__appmap_disabled(self):
```
apotterri, line 17: I think it would be better to use classes to scope test cases, rather than a naming convention.

appmap/test/test_detect_enabled.py
symwell: I don't know.  I'm seeing cases where `monkeypatch.setenv("APPMAP_RECORD_REQUESTS", "false")` doesn't do what it should: the variable doesn't get added in os.environ.

apotterri: That seems very strange. Maybe something else is going on? Are you using `monkeypatch.setenv` everywhere, or are there still changes be made directly to `os.environ`?

symwell: I think the variable was being added but by the time my code got to execute it was no longer there, as if something had cleaned `os.environ`.

appmap/test/test_detect_enabled.py
```diff
@@ -0,0 +1,194 @@
...
+"""
+            + extra_settings_content
+            + """
+"""
+        )
+        monkeypatch.syspath_prepend(data_dir / "django")
+        monkeypatch.setenv("DJANGO_SETTINGS_MODULE", django_settings_module)
+        filename = data_dir / "django" / "app" / basename_settings
+        try:
+            file_write(str(filename), settings_content)
```
apotterri, line 148: You should use a `with` statement for this, and get rid of `file_write`:
https://github.com/getappmap/appmap-python/blob/5858681742e47e6841ecd32379c297fded9df4ef/appmap/_implementation/configuration.py#L234

appmap/_implementation/detect_enabled.py
```diff
@@ -0,0 +1,182 @@
...
+        if self._initialized:
+            return
+
+        self._initialized = True
+        self._detected_for_method = {}
+
+    @classmethod
+    def initialize(cls):
+        cls._instance = None
+        # because apparently __new__ and __init__ don't get called
```
apotterri, line 45: Right, because you're not actually asking to create an instance, i.e. you don't say `DetectEnabled()` anywhere.

appmap/_implementation/detect_enabled.py
```diff
@@ -0,0 +1,182 @@
...
+
+import importlib
+import logging
+import os
+from textwrap import dedent
+
+logger = logging.getLogger(__name__)
+
+
+class UnrecognizedRecordingMethodException(Exception):
```
apotterri, line 13: Doesn't look like this is getting used anywhere?

appmap/django.py
```diff
@@ -211,19 +212,22 @@ def normalize_path_info(path_info, resolved):
...
+        super().__init__()
         self.get_response = get_response
         self.recorder = Recorder()
-        self.record_url = "/_appmap/record"
 
     def __call__(self, request):
-        if not Env.current.enabled:
+        if not (
+            DetectEnabled.should_enable("remote")
+            or DetectEnabled.should_enable("requests")
```
apotterri, line 222: You could DRY up the code in the individual frameworks by adding a method to AppmapMiddleware like
```python
    @classmethod
    def should_record(cls):
        return DetectEnabled.should_enable("remote")
        or DetectEnabled.should_enable("requests")
```
then saying
```python
    if not self.should_record():
```
in the subclasses.

appmap/test/test_detect_enabled.py
apotterri: Hmmm. You marked this as resolved, but didn't change the naming convention? I guess that's your call, but maybe you could explain?

apotterri (changes requested):
As I mentioned in one of my comments, please use one or more `fixup` commits when you make changes after a review. It'll make it much easier see the new code.

appmap/test/test_detect_enabled.py
symwell: Done.  I would have kept `file_write`.  A function that writes in a file seems like a useful abstraction.  I'm surprised there's still no such function in Python while there is one in Ruby.

appmap/_implementation/detect_enabled.py
apotterri: It looks to me like it gets set. Maybe I don't understand what you mean?

```
diff --git a/appmap/_implementation/detect_enabled.py b/appmap/_implementation/detect_enabled.py
index 6d03c7b..aeea80c 100644
--- a/appmap/_implementation/detect_enabled.py
+++ b/appmap/_implementation/detect_enabled.py
@@ -38,11 +38,12 @@ class DetectEnabled:
     def initialize(cls):
         cls._instance = None
         # because apparently __new__ and __init__ don't get called
-        cls._detected_for_method = {}
+        # cls._detected_for_method = {}

     @classmethod
     def clear_cache(cls):
-        cls._detected_for_method = {}
+        # cls._detected_for_method = {}
+        pass

     @classmethod
     def is_appmap_repo(cls):
sw/feat/record_by_default *$
ajp@Alans-MacBook-Pro appmap-python-2 % python
Python 3.10.7 (main, Sep 15 2022, 04:53:36) [Clang 13.1.6 (clang-1316.0.21.2.5)] on darwin
>>> from appmap._implementation.detect_enabled import DetectEnabled
>>> de = DetectEnabled()
>>> print(de._detected_for_method)
{}
>>>
```

symwell: If I apply the first diff -- if I comment out `cls._detected_for_method = {}` in `initialize` -- the Flask testcases don't pass
```
poetry run pytest appmap/test/test_flask.py
```

appmap/_implementation/detect_enabled.py
```diff
@@ -0,0 +1,182 @@
...
+                        return [
+                            f"{os.environ['DJANGO_SETTINGS_MODULE']}.DEBUG={settings.DEBUG}",
+                            True,
+                        ]
+                    elif settings.DEBUG == False:
+                        return [
+                            f"{os.environ['DJANGO_SETTINGS_MODULE']}.DEBUG={settings.DEBUG}",
+                            False,
+                        ]
+                except Exception as exn:
```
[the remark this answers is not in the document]
apotterri: So, then the code should explicitly check for just that. It can do that be either catching *the specific exception* that gets raised in that case, or check to see if `settings` has properties named `DEBUG` or `APPMAP`. Catching `Exception` is too broad, and leads to bugs like the one I originally called out.

appmap/_implementation/detect_enabled.py
```diff
@@ -0,0 +1,182 @@
...
+        # because apparently __new__ and __init__ don't get called
+        cls._detected_for_method = {}
+
+    @classmethod
+    def clear_cache(cls):
+        cls._detected_for_method = {}
+
+    @classmethod
+    def is_appmap_repo(cls):
+        return os.path.exists("appmap/_implementation/event.py") and os.path.exists(
```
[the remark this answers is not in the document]
apotterri: I still don't love this solution. But, if we have to go this way, let's at least test for a file that's less likely to move around if we refactor, e.g. `appmap/__init__.py`, or `appmap/_implementation/__init__.py`.

appmap/_implementation/detect_enabled.py
apotterri: Well, ok, but that's not (necessarily) the same thing as "__init__ doesn't set a property of self", which is what it sounds like you meant.

appmap/_implementation/detect_enabled.py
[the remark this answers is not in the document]
symwell: I see.  Yeah, it should catch a specific exception instead of any exception that comes up.  Fixed.

apotterri (commented):
Somehow, GitHub decided that some of my comments should be one-offs, and others should be part of a review. So, I'm just submitting this as "general feedback"....

symwell:
This is ready to review.  The build passes now.

Had to change a lot of testcases after the recent rebase.

appmap/flask.py
```diff
@@ -47,21 +48,14 @@ def request_params(req):
 
 class AppmapFlask(AppmapMiddleware):
     def __init__(self, app=None):
+        super().__init__()
         self.app = app
         if app is not None:
             self.init_app(app)
 
     def init_app(self, app):
-        if not Env.current.enabled:
```
apotterri, line 55: Removing all these checks means that even if recording is disabled (e.g. with `APPMAP=false`), the remote-recording routes will still be added to the app, right? Is there some reason that's necessary?

apotterri (commented):
One more issue to resolve, then this should be ready for a final rebase.

appmap/flask.py
symwell: That's right, the remote-recording routes will still be added.

Without removing this check some testcase wasn't working.  I think it was [test_starts_disabled](https://github.com/getappmap/appmap-python/blob/master/appmap/test/web_framework.py#L235)

I wish my commit history wasn't squashed into a single commit so that I could find through the commit comments which testcase this was.

apotterri: I don't think we want the routes added if recording is disabled. To fix `test_starts_disabled`, couldn't you instead make sure that remote recording is enabled (e.g. with `APPMAP=true`?

symwell: This was before the rebase of the events pr, so let me try putting it back to see what I get.

symwell: It was failing two other testcases, not the one I said:
```
FAILED appmap/test/test_flask.py::TestRecordRequestsFlask::test_record_request_appmap_not_enabled_requests_enabled_and_remote - AssertionError
FAILED appmap/test/test_flask.py::TestRecordRequestsFlask::test_record_request_appmap_not_enabled_requests_enabled_no_remote - KeyError: 'appmap-file-name'
```

Also, it was failing this testcases only for Flask.  Apparently there was no such check for the routes for Django.

I think it's because I was trying to test the case of `APPMAP=false` (appmap not enabled) but then enabling remote recording through the testcase issuing a POST.

My understanding is that remote recording and `APPMAP=true` are separate features.  Is this true, or is remote recording possible only if `APPMAP=true`?

apotterri: > Is this true, or is remote recording possible only if APPMAP=true?
Remote recording is only possible if APPMAP=true.

apotterri: > Apparently there was no such check for the routes for Django.

This seems like a bug.

apotterri:
> I wish my commit history wasn't squashed into a single commit so that I could find through the commit comments which testcase this was.

That seems reasonable. It's basically the workflow described here: https://git-scm.com/docs/git-rebase#_interactive_mode, amended for GitHub PRs:

1. have a wonderful idea

2. hack on the code

3. prepare a series for submission, request a review

    a. apply `fixup!` commits to address review comments
  
    b. re-request review

    c. repeat a and b as necessary

4. final history cleanup, squashing original commits into one or two final, orthogonal commits

My experience has been that step 4 can be somewhat challenging, if steps 1-3 have touched many files. Sometimes you can avoid this by making the initial task smaller, but not always.

appmap/flask.py
symwell: I think something's wrong.  In [test_can_record](https://github.com/getappmap/appmap-python/blob/master/appmap/test/web_framework.py#L263) I'm seeing remote requests enabled with a POST but then by the time the DELETE is issued to disable them recording is already off.  As if some `.stop_recording()` call disabled recording.

apotterri (approved):

PR #177, opened 2022-10-05
Fix per-thread event support

apotterri:
**I left the `refactor` commits separate to make this easier to review, but I think they should be squashed before this gets merged.**

Fixes getappmap/board#184.


appmap/_implementation/recorder.py
```diff
@@ -1,81 +1,184 @@
...
+            return super().next_event_id
+
+    def _add_event(self, event):
+        with self._lock:
+            super()._add_event(event)
 
 
 def initialize():
-    Recorder.initialize()
+    Recorder._initialize()
```
dividedmind, line 184: It seems to me it doesn't do any initialization anymore that would require other things set up before. Should it just be created at top-level statically? Eg.
```python
_default_recorder = SharedRecorder()
```

This way you could get rid of all the ugly `initialize()` methods.

appmap/_implementation/recorder.py
```diff
@@ -1,81 +1,184 @@
...
+    @abstractmethod
+    def next_event_id(self):
+        self._next_event_id += 1
+        return self._next_event_id
+
+    # It might be nice to put @property on the getters here. The python maintainers have gone back
+    # and forth on whether you should be able to combine @classmethod and @property. In 3.11,
+    # they've decided you can't: https://docs.python.org/3.11/library/functions.html#classmethod.
+    @classmethod
+    def get_current(cls):
```
dividedmind, line 35: How about making this `__new__` instead? This way you could do `Recorder()` instead of `Recorder.get_current()`. Or would it be too confusing? :)

appmap/_implementation/recorder.py
```diff
@@ -1,81 +1,184 @@
...
+
+    Note that the abstract methods have implementations for use by subclasses.
+    """
+
+    @property
+    @abstractmethod
+    def events(self):
+        return self._events
+
+    @property
```
dividedmind, line 25: I don't think this should be a `@property`; I wouldn't expect a property access to mutate state (except for caching or reification).

dividedmind (changes requested):
Thanks! But there is a silent functional change here: only one recorder can be active at the same time now. This is not only different from how appmap-ruby works, it can also be confusing. If I have streaming recording enabled and try to use remote recording, the remote recording will unexpectedly come back (mostly) empty.

dividedmind:
> I left the `refactor` commits separate to make this easier to review, but I think they should be squashed before this gets merged.

BTW, thanks for this, it does make it much easier to review :)

apotterri:
> Thanks! But there is a silent functional change here: only one recorder can be active at the same time now. This is not only different from how appmap-ruby works, it can also be confusing. If I have streaming recording enabled and try to use remote recording, the remote recording will unexpectedly come back (mostly) empty.

Thanks for pointing this out. I went back and forth about how to handle this situation, and didn't really see a reason to support both types of recording. I'm persuaded by your argument, though, so I'll make it work like appmap-ruby.

appmap/_implementation/recorder.py
apotterri: The main reason for `initialize` is to reset the state of all the modules before each test case runs. I think that's still necessary, and I don't really want to take the time to investigate it for this PR.

appmap/_implementation/recorder.py
```diff
@@ -0,0 +1,184 @@
...
+    @abstractmethod
+    def next_event_id(self):
+        self._next_event_id += 1
+        return self._next_event_id
```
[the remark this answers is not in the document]
apotterri: There's no need for a method decorated with `@abstractmethod` to throw an exception . Inheriting from `ABC` ensures that only concrete subclasses can be instantiated. 

I think it's important to force the choice between a `ThreadRecorder` and a `SharedRecorder`. However, the since the difference between the two is only the locking required to ensure that `SharedRecorder` is thread-safe, having them share implementations seems like the right way to go.

Both of these points are discussed here: https://docs.python.org/3/library/abc.html?highlight=abc#abc.abstractmethod .

appmap/_implementation/recorder.py
apotterri: Done.

symwell:
I wonder if someday we'll want to have three recorders recording at the same time instead of at most two.

appmap/__init__.py
```diff
@@ -2,7 +2,8 @@
 from ._implementation import generation  # noqa: F401
 from ._implementation.env import Env  # noqa: F401
 from ._implementation.labels import labels  # noqa: F401
-from ._implementation.recording import Recording, instrument_module  # noqa: F401
+from ._implementation.importer import instrument_module  # noqa: F401
```
symwell, line 5: Is `from ._implementation.importer import instrument_module` necessary?

symwell (approved):

Issue #163, opened 2022-07-27
AppMap agent for Python fails to install when PyYAML version conflicts


dustinbyrne:
Is is possible to relax the constraints on this dependency?

```
ERROR: Cannot install -r requirements.txt (line 1) and PyYAML==6.0 because these package versions have conflicting dependencies.

The conflict is caused by:
The user requested PyYAML==6.0
appmap 1.5.1 depends on PyYAML<6.0.0 and >=5.3.0
```

symwell:
I ran into this problem too, after adding `appmap==1.10.0` in saleor-platform's requirements.txt

PR #179, opened 2022-10-13
fix: Require PyYAML >=5.3.0 rather than ^5.3.0

symwell:
Some packages that use appmap-python, like saleor, require PyYAML==6.0 and this conflicts with appmap-python requiring 5.3.0.

Fixes https://github.com/getappmap/appmap-python/issues/163


apotterri (approved):
LGTM

PR #180, opened 2022-10-17
Fix record-by-defauilt

apotterri:
Make sure it instruments application code. There's definitely more testing that could be added here, but [this](https://github.com/getappmap/appmap-python/pull/180/commits/ce0b5239b13608db2874bed175d193267af4ce82#diff-f0da5d71b031834061fcaa37daf4a32c38ab2e6030993abdcbb8e0d7036c22f5R112) is a start.

Fixes getappmap/board#191.

### DIFF
--- a/appmap/flask.py
+++ b/appmap/flask.py
@@ -57,8 +57,7 @@ class AppmapFlask(AppmapMiddleware):
             # it may record requests but not remote (APPMAP=false)
             self.recorder = Recorder.get_current()
 
-        if Env.current.enabled:
-            # the remote recording routes are enabled only if APPMAP=true
+        if DetectEnabled.should_enable("remote"):
             app.add_url_rule(
                 self.record_url,
                 "appmap_record_get",
--- a/appmap/test/test_configuration.py
+++ b/appmap/test/test_configuration.py
@@ -1,7 +1,6 @@
 """Test Configuration"""
 # pylint: disable=missing-function-docstring
 
-import sys
 from distutils.dir_util import copy_tree
 from pathlib import Path
 
@@ -9,7 +8,6 @@ import pytest
 
 import appmap
 import appmap._implementation
-import appmap._implementation.env as impl_env
 from appmap._implementation.configuration import Config, ConfigFilter
 from appmap._implementation.env import Env
 from appmap._implementation.importer import Filterable, NullFilter
@@ -47,6 +45,7 @@ def test_reports_invalid():
     assert not Config().file_valid
 
 
+@pytest.mark.appmap_enabled(config="appmap-broken.yml", appmap_enabled=None)
 def test_is_disabled_when_unset():
     """Test that recording is disabled when APPMAP is unset"""
     assert Env.current.get("APPMAP", None) is None
@@ -54,18 +53,10 @@ def test_is_disabled_when_unset():
     assert not appmap.enabled()
 
 
+@pytest.mark.appmap_enabled(config="appmap-broken.yml", appmap_enabled="false")
 def test_is_disabled_when_false():
     """Test that recording is disabled when APPMAP=false"""
     Env.current.set("APPMAP", "false")
-    assert not appmap.enabled()
-
-
-@pytest.mark.appmap_enabled(appmap_enabled=None)
-def test_is_disabled_with_valid_config():
-    c = Config()
-    assert c.file_present
-    assert c.file_valid
-
     assert not appmap.enabled()
 
 
@@ -86,15 +77,16 @@ def test_config_not_found(caplog):
     assert f'"{not_found}" is missing' in caplog.text
 
 
+@pytest.mark.appmap_enabled(appmap_enabled="false", config="notfound.yml")
 def test_config_no_message(caplog):
     """
     Messages about a missing config should only be logged when
     recording is enabled
     """
 
-    assert Config().name is None
     assert not appmap.enabled()
-    assert caplog.text is ""
+    assert Config().name is None
+    assert caplog.text == ""
 
 
 cf = lambda: ConfigFilter(NullFilter())
@@ -200,6 +192,7 @@ class TestDefaultConfig:
         assert path.is_file()
 
     def test_not_created_if_missing_and_not_enabled(self, git, data_dir, monkeypatch):
+        monkeypatch.setenv("APPMAP", "false")
         repo_root = git.cwd
         copy_tree(data_dir / "config", str(repo_root))
         monkeypatch.chdir(repo_root)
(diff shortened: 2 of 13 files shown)


appmap/test/web_framework.py
```diff
@@ -363,6 +369,10 @@ def server_url():
 
     @staticmethod
     def record_request_thread():
+        # I've seen occasional test failures, seemingly because the test servers can't handle the
+        # barrage of requests. A tiny bit of delay still causes many, many concurrent requests, but
+        # eliminates the failures.
+        time.sleep(SR.uniform(0, 0.1))
```
symwell, line 375: I ran `poetry run pytest -s -v appmap/test/test_flask.py` while also running saleor tests and got:
```
            # wait for all threads to complete
            for future in concurrent.futures.as_completed(future_to_request_number):
                try:
                    response = future.result()
                except Exception as e:
                    traceback.print_exc()
>               assert response.status_code == 200
E               assert 400 == 200
E                +  where 400 = <Response [400]>.status_code

appmap/test/web_framework.py:410: AssertionError
```

Also ran when not running anything else and got:

```
FAILED appmap/test/test_flask.py::TestRecordRequestsFlask::test_record_request_appmap_enabled_requests_enabled_no_remote - assert 400 == 200
FAILED appmap/test/test_flask.py::TestRecordRequestsFlask::test_record_request_appmap_enabled_requests_enabled_and_remote - assert 400 == 200
FAILED appmap/test/test_flask.py::TestRecordRequestsFlask::test_record_request_appmap_not_enabled_requests_enabled_no_remote - assert 400 == 200
```

Let's keep this change though.  I can imagine it helps. The real problem may be the ability of the tested webserver (Flask/Django) to handle concurrency.

appmap/flask.py
```diff
@@ -113,16 +117,15 @@ def before_request(self):
...
-            np = "".join(
-                [
-                    f"{{{p}}}" if c else p
-                    for c, _, p in parse_rule(request.url_rule.rule)
-                ]
+            # Transform requestl.url to the expected normalized-path form. For example,
+            # "/post/<username>/<post_id>/summary" becomes "/post/{{username}}/{{post_id}}/summary".
+            r = repr(request.url_rule)
+            np = NP_XLATE.sub(
+                lambda m: "{" if m["l"] else "}" if m["r"] else "",
```
dividedmind, line 125: Should this be double `{{` and `}}` or is the comment inaccurate? Regardless, the comment could show the `repr` format so it's clear what's happening.

(I'm kind of weirded out by using `repr` like this, but as long as it works... also, let's hope rules never have a `'` in the name.)

appmap/flask.py
```diff
@@ -113,16 +117,15 @@ def before_request(self):
...
+                NP_REGEXP.findall(r)[0],
```
dividedmind, line 126: I don't think this works. I tried `r = '/post/<username>/<post_id>/summary'` — I got `IndexError: list index out of range`. I tried `r = '/post/<Rule \'username\'>/<Rule \'post_id\'>/summary'` — I got `np = 'username'`. Do we have tests for this?

Also, does this even belong in this pull request?

appmap/test/test_configuration.py
```diff
@@ -47,19 +45,24 @@ def test_reports_invalid():
...
+@pytest.mark.appmap_enabled(config="appmap-broken.yml", appmap_enabled="false")
 def test_is_disabled_when_false():
     """Test that recording is disabled when APPMAP=false"""
     Env.current.set("APPMAP", "false")
     assert not appmap.enabled()
 
 
+@pytest.mark.skip(
+    reason="TODO: determine if this test makes sense, appmap is enabled by default"
+)
```
dividedmind, line 65: Yeah, should probably just remove it.

dividedmind (changes requested):
The detect enabled part seems fine. But there seems to be a stray broken flask commit there.

symwell:
Is it worth adding a testcase that expects the importer to work `if Env.current.enabled`?

appmap/flask.py
apotterri: The comment is wrong. I'll add the `repr` format.

Using `repr` feels better to me than reaching into the implementation and using an undocumented interface (which was completely rewritten for Flask 2). The `repr` format is the same for both versions, so it should work for a least a little while....

appmap/flask.py
apotterri: Yes, there's a test for it, but I'll back and look at what's going on with.

Well, including it seemed harmless, but clearly it's not working well enough to keep. I'll take it out, and fix whatever problems it has as part of the Flask 2 work.

appmap/test/test_configuration.py
apotterri: WIll do.

apotterri:
> Is it worth adding a testcase that expects importer.py to work if Env.current.enabled?

If you're asking for a test case that confirms that application code gets instrumented for record-by-default, that's effectively what deleting `APPMAP` from the environment does here: https://github.com/getappmap/appmap-python/blob/ce0b5239b13608db2874bed175d193267af4ce82/appmap/test/test_test_frameworks.py#L112 (this was the change the link in my original comment was supposed to point to, sorry about that). I went back and confirmed that if I made that change to the code on `master`, test failed because application code didn't appear in the generated recordings.

If that's not what you're asking for, though, maybe you could clarify a bit? Thanks!

appmap/test/web_framework.py
symwell: The failures I saw turned out to be due to a saleor container using the same port number, 8000.
All tests pass after I stopped this container.  All good.

symwell:
> If that's not what you're asking for, though, maybe you could clarify a bit?

I might not fully understand how appmap gets enabled.

I thought appmap gets enabled in two ways
- from the code in `appmap/__init__.py` installed through pip.  This part works.  When a standalone Python program in a single file runs, all the functions in it are instrumented.
- when `import some_module_name` is used by a program.  I thought something more needs to happen for appmap to instrument code for this import.  I thought there would be a testcase in two standalone files where the first tries to import the second.

appmap/flask.py
```diff
@@ -57,7 +57,7 @@ def init_app(self, app):
             # it may record requests but not remote (APPMAP=false)
             self.recorder = Recorder.get_current()
 
-        if Env.current.enabled:
+        if DetectEnabled.should_enable("remote"):
             # the remote recording routes are enabled only if APPMAP=true
```
dividedmind, line 61: The comment is stale now.

dividedmind (approved):
Looks good, thanks :)

appmap/flask.py
apotterri: Thanks, fixed.

PR #184, opened 2022-10-20
fix: use default values when config is incomplete

### LINKED ISSUE #181: AppMap agent introduces test failures in pytest
```
WARNING appmap._implementation.detect_enabled: AppMap recording is enabled because will record by default
WARNING:appmap._implementation.detect_enabled:AppMap recording is enabled because will record by default

...
self = <appmap._implementation.configuration.Config object at 0x105a43388>

@property
def name(self):
>       return self._config["name"]
E KeyError: 'name'

appmap/_implementation/configuration.py:159: KeyError
```

apotterri:
If the config file is missing a value for `name` or `packages`, use the default value instead.

Fixes #181

### DIFF
--- a/appmap.pth
+++ b/appmap.pth
@@ -1,1 +1,2 @@
-import appmap
+# The python runtime suppresses errors from import, but we should fail if something goes wrong. 
+try: import appmap; except Exception: raise
--- a/appmap/_implementation/configuration.py
+++ b/appmap/_implementation/configuration.py
@@ -173,11 +173,17 @@ class Config:
 
     @property
     def default(self):
+        return {"name": self.default_name, "packages": self.default_packages}
+
+    @property
+    def default_name(self):
         root_dir = Env.current.root_dir
-        return {
-            "name": default_app_name(root_dir),
-            "packages": [{"path": p} for p in find_top_packages(root_dir)],
-        }
+        return default_app_name(root_dir)
+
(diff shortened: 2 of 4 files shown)


appmap.pth
```diff
@@ -1 +1 @@
-import appmap
\ No newline at end of file
+try: import appmap; except Exception: raise
```
symwell, line 1: How does this change program behavior?  Wouldn't an exception have been raised before if `import appmap` failed?

symwell (approved):

appmap.pth
apotterri: The behavior here is a little strange. Without this change, the python runtime displays a traceback showing where the import failed, but allows the script to continue. This leads to builds passing that should fail, e.g. https://app.travis-ci.com/github/getappmap/appmap-python/builds/256844492#L2780 . In this case, `appmap-agent-init` writes the expected config JSON, even though the import failed.

With the `try ... except: raise` change, `appmap-agent-init` no longer writes the expected JSON, the build fails, and we know we have a problem.

In practice, given that we shouldn't be deploying broken builds, I don't expect the user to ever see the result of this change.

symwell: I suggest adding a comment there with this explanation.

apotterri: Done.

PR #186, opened 2022-10-25
Add Flask 2 support

apotterri:
Fixes getappmap/board/#148.


appmap/flask.py
```diff
@@ -45,62 +44,40 @@ def request_params(req):
...
+    """
+
+    def __init__(self):
         super().__init__()
-        self.app = app
-        if app is not None:
-            self.init_app(app)
 
     def init_app(self, app):
         if self.should_record():
```
symwell, line 70: The following check used to be in `init_app` and I had [removed](https://github.com/getappmap/appmap-python/pull/176/files#diff-144faf942a3ddaa37c23c0c4e9c67f3912c28aa7e267466a2250f4947be0431cL58) it.  Since `init_app` can now be called by the user, this check should probably be added back.

```
if not app:
    return
```

Maybe it should `raise` to let the user know app was `None`.

appmap/test/data/django/app/settings.py
```diff
@@ -10,5 +10,3 @@
 # Must set ROOT_URLCONF else we get
 # AttributeError: 'Settings' object has no attribute 'ROOT_URLCONF'
 ROOT_URLCONF = "app.urls"
-
-MIDDLEWARE = ["appmap.django.Middleware"]
```
symwell, line 14: Why was this removed?  Is it because it's expected that appmap will change this variable to insert `"appmap.django.Middleware"` in it ?

apotterri:
@symwell It's possible that [c474ff9](https://github.com/getappmap/appmap-python/pull/186/commits/c474ff93dfc875b6a43eb7dc66f371383d9e1cf2) fixes the issues with record-by-default in [chypi.org](https://github.com/chicagopython/chipy.org).

appmap/flask.py
apotterri: It doesn't ever make sense to call `init_app` with `None`. Does it really add anything to check `app` and raise if it's `None`, rather than just allowing the runtime to do the check?

symwell: Allowing the runtime to do the check seems better.

appmap/test/data/django/app/settings.py
apotterri: Yes, exactly. By not adding the middleware manually, the tests that use this app confirm that it's automatically injected.

symwell (approved):

appmap/flask.py
```diff
@@ -112,16 +116,20 @@ def before_request(self):
...
         if request.url_rule:
-            np = "".join(
-                [
-                    f"{{{p}}}" if c else p
-                    for c, _, p in parse_rule(request.url_rule.rule)
-                ]
+            # Transform request.url to the expected normalized-path form. For example,
+            # "/post/<username>/<post_id>/summary" becomes "/post/{username}/{post_id}/summary".
+            # Notes:
+            #   * the value of `repr` of this rule begins with "<Repr '/post/<username>/<post_id>/summary'"
```
dividedmind, line 96: Should this say `Rule`?

appmap/flask.py
```diff
@@ -112,16 +116,20 @@ def before_request(self):
...
+            #   * the variable names in a rule can only contain alphanumerics:
+            #     * flask 1: https://github.com/pallets/werkzeug/blob/1dde4b1790f9c46b7122bb8225e6b48a5b22a615/src/werkzeug/routing.py#L143
+            #     * flask 2: https://github.com/pallets/werkzeug/blob/99f328cf2721e913bd8a3128a9cdd95ca97c334c/src/werkzeug/routing/rules.py#L56
+            r = repr(request.url_rule)
+            np = NP_XLATE.sub(
+                lambda m: "{" if m["l"] else "}" if m["r"] else "",
```
dividedmind, line 102: That's a bit complicated a way to do this simple translation. Did you consider using `translate`, eg. `re.compile("'(.*)'").findall(repr(request.url_rule))[0].translate(str.maketrans('<>', '{}')` ?

appmap/flask.py
```diff
@@ -112,16 +116,20 @@ def before_request(self):
...
-        # See
-        # https://github.com/pallets/werkzeug/blob/2.0.0/src/werkzeug/routing.py#L213
-        # for a description of parse_rule.
         if request.url_rule:
-            np = "".join(
-                [
-                    f"{{{p}}}" if c else p
-                    for c, _, p in parse_rule(request.url_rule.rule)
-                ]
+            # Transform request.url to the expected normalized-path form. For example,
```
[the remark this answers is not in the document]
dividedmind: Thanks, looks ok if a bit convoluted :) I'll propose a simpler and perhaps easier to read implementation

dividedmind (approved):

appmap/flask.py
apotterri: Yup, fixed.

appmap/flask.py
apotterri: Nope, didn't know about `translate`. I agree that's better.

PR #190, opened 2022-11-05
Remote and requests 20221026

apotterri:
With these changes, remote recording and per-request recording no longer
interfere with one another.

Fixes getappmap/board#203.

### DIFF
--- a/appmap/_implementation/recorder.py
+++ b/appmap/_implementation/recorder.py
@@ -2,6 +2,7 @@ import logging
 import threading
 import traceback
 from abc import ABC, abstractmethod
+from copy import copy
 
 from appmap._implementation.utils import appmap_tls
 
@@ -22,10 +23,14 @@ class Recorder(ABC):
     def events(self):
         return self._events
 
-    @abstractmethod
-    def next_event_id(self):
-        self._next_event_id += 1
-        return self._next_event_id
+    _next_event_id = 0
+    _next_event_id_lock = threading.Lock()
+
+    @classmethod
+    def next_event_id(cls):
+        with cls._next_event_id_lock:
+            cls._next_event_id += 1
+            return cls._next_event_id
 
     # It might be nice to put @property on the getters here. The python maintainers have gone back
     # and forth on whether you should be able to combine @classmethod and @property. In 3.11,
@@ -44,8 +49,24 @@ class Recorder(ABC):
         """
         Set the recorder for the current thread.
         """
-        appmap_tls()[cls._RECORDER_KEY] = r
+        tls = appmap_tls()
+        if r:
+            tls[cls._RECORDER_KEY] = r
+        else:
+            del tls[cls._RECORDER_KEY]
+
         return r
+
+    @classmethod
+    def get_global(cls):
+        _, shared = cls._get_current()
+        return shared
+
+    @classmethod
+    def new_global(cls):
+        global _default_recorder
+        _default_recorder = SharedRecorder()
+        return _default_recorder
 
     @classmethod
     def get_enabled(cls):
@@ -85,13 +106,11 @@ class Recorder(ABC):
 
     def clear(self):
         self._events = []
-        self._next_event_id = 0
 
     def __init__(self, enabled=False):
         self._events = []
         self._enabled = enabled
         self.start_tb = None
-        self._next_event_id = 0
 
     @abstractmethod
     def _start_recording(self):
@@ -122,8 +141,7 @@ class Recorder(ABC):
         threads initializing the default recorder. If you find yourself wanting to do that, you
         should probably be using per-thread recording.
         """
-        global _default_recorder
-        _default_recorder = SharedRecorder()
+        Recorder.new_global()
 
 
 class ThreadRecorder(Recorder):
@@ -134,9 +152,6 @@ class ThreadRecorder(Recorder):
     @property
     def events(self):
         return super().events
-
-    def next_event_id(self):
-        return super().next_event_id()
 
     def _start_recording(self):
         super()._start_recording()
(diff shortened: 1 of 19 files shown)


appmap/_implementation/recorder.py
```diff
@@ -147,13 +163,23 @@ def _stop_recording(self):
...
+    # pragma pylint: enable=useless-super-delegation
+
 
 class SharedRecorder(Recorder):
     """
     A shared Recorder. The global recorder is an instance of this class.
     """
 
-    _lock = threading.Lock()
+    _lock = threading.RLock()
```
symwell, line 174: It may be worth explaining in a comment how the non re-entrancy lead to what this suggests to have been a deadlock.  How did a thread, after acquiring this lock, called code that tried to acquire the lock again?

appmap/test/test_django.py
```diff
@@ -235,7 +230,11 @@ def server_start(env_vars_str):
     @staticmethod
     def server_stop():
         exec_cmd(
-            "ps -ef | grep -i 'manage.py runserver' | grep -v grep | awk '{ print $2 }' | xargs kill -9"
+            "ps -ef"
+            + "| grep -i 'manage.py runserver'"
+            + "| grep -v grep"
+            + "| awk '{ print $2 }'"
+            + "| xargs kill -9"
```
symwell, line 237: Nice.  May as well do the same in `test_flask.py`.

per-thread.md
```diff
@@ -0,0 +1,26 @@
...
+      global recorder
+
+Global recording app:
+[✓] Flask
+[ ] Django
+* Is installed if should_enable("remote"), no other decisions should be made based on
+  should_enabled("remote")
+* Manages the state of remote recording separately from Recoder.get_current().enabled, so...
+* Doesn't make any decisions based on Recorder.get_current().enabled
+
```
symwell, line 11: FYI the checkmarks were displayed horizontally rather than vertically.

checkmarks

appmap/_implementation/recorder.py
```diff
@@ -22,10 +23,14 @@ class Recorder(ABC):
...
-        self._next_event_id += 1
-        return self._next_event_id
+    _next_event_id = 0
+    _next_event_id_lock = threading.Lock()
+
+    @classmethod
+    def next_event_id(cls):
+        with cls._next_event_id_lock:
+            cls._next_event_id += 1
+            return cls._next_event_id
```
symwell, line 33: What was the interference of remote requests with remote recording?

Was part of it a race condition over event ids?  If event ids are unique per thread then why put them behind a lock?

symwell (commented):
Lots of good stuff for appmap-python!  A couple of questions to understand the design better.

per-thread.md
apotterri: Oops, these were just some notes for myself. I'll delete this file.

appmap/_implementation/recorder.py
apotterri: Good catch. Using an `RLock` here was a holdover from a previous attempt to fix this problem. It's no longer necessary, so I've switched back to `Lock`.

appmap/test/test_django.py
apotterri: Done.

per-thread.md
apotterri: Done.

appmap/_implementation/recorder.py
apotterri: There were three two aspects to the interference:
* remote recording and per-request recording were both trying to use Recorder.get_current().enabled to manage the state of remote recording
* new event ids weren't getting issued properly

I fixed the first by tracking the state of remote recording with a `ContextVar` with a boolean value. I also simplified the code somewhat by separating it from the middleware that captures `http_server` events.

I fixed the second by having the next event id be global (it's a class variable of `Recorder`, which is why there's locking around updating it). Careful management of resetting it meant that no tests need to change. That management probably also means it's a bit fragile, but perhaps that an issue for another time.

One thing that might be surprising about this PR: there aren't any updates for the expected AppMaps. This is actually because of the problem identified with AppMap correctness: the only route in the test Flask app is implemented as a [function in module](https://github.com/getappmap/appmap-python/blob/bb33143582c68433d59800fd8e1f050bd4ba36f1/appmap/test/data/flask/app.py#L15) , so it doesn't show up in the recordings.

WDYT, does it seem worth adding a test that show enabling both kinds of recording results in a proper AppMap?

symwell: I think
- this pr is worth shipping as is
- it's worth adding a low priority issue to revisit event ids.  If event ids are unique per thread then it appears possible to not need a global lock for event ids.  At this point this looks like an optimization, so it doesn't appear like a priority.

[test_record_request_appmap_enabled_requests_enabled_and_remote](https://github.com/getappmap/appmap-python/blob/master/appmap/test/test_django.py#L251) collects both kinds of recording results.  I wonder if at least one of the many threads there contains data collected for remote recording.  Maybe all it needs is an assert or two in [record_request_thread](https://github.com/getappmap/appmap-python/blob/master/appmap/test/web_framework.py#L371).

symwell (approved):

appmap/_implementation/recorder.py
symwell: This doesn't work for Django for me:
```
APPMAP=true APPMAP_RECORD_REQUESTS=true APPMAP_OUTPUT_DIR=/tmp python3 manage.py runserver 0.0.0.0:8000

curl -s localhost:8000/_appmap/record
```
The `_appmap/record` doesn't exist.

symwell (changes requested):
Django/Flask `_appmap/record` endpoints don't work for me.

appmap/_implementation/recorder.py
apotterri: Try running it like this:
```
PYTHONPATH=init APPMAP=true APPMAP_RECORD_REQUESTS=true APPMAP_OUTPUT_DIR=/tmp  FLASK_DEBUG=1 FLASK_APP=app.py flask run -p 8000
```
Setting `PYTHONPATH` here simulates what happens when the package is really installed.

symwell (approved):

PR #191, opened 2022-11-08
fix: handle a tuple value for settings.MIDDLEWARE

apotterri:
Django is fine with settings.MIDDLEWARE being either a tuple or a list, but we were expecting a list. Now, we can handle either.

Fixes getappmap/board#225.

### DIFF
--- a/appmap/django.py
+++ b/appmap/django.py
@@ -6,6 +6,7 @@ and run with APPMAP=true set in the environment.
 """
 
 import json
+import logging
 import re
 import sys
 import time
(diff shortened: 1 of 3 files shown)


appmap/django.py
```diff
@@ -296,10 +299,23 @@ def after_request_main(self, rec, response, start, call_event_id):
...
+        # Django is ok with settings.MIDDLEWARE being any kind iterable. Update it, without changing
+        # its type, if we can.
+        if isinstance(settings.MIDDLEWARE, list):
+            settings.MIDDLEWARE = stack
+        elif isinstance(settings.MIDDLEWARE, tuple):
+            settings.MIDDLEWARE = tuple(stack)
+        else:
+            logger.warning(
+                f"Don't know how to update settings.MIDDLEWARE of type {type(settings.MIDDLEWARE)}, recording is not enabled."
+            )
```
symwell, line 318: Nice!

symwell (approved):

cktricky:
This is awesome! Nice patch 🙇 ✨

PR #193, opened 2022-11-08
Modify the AST to instrument functions before they're loaded.

apotterri:


apotterri:
The introduction of `sys.monitoring` in 3.12 means we (almost certainly) won't need to do this.

PR #202, opened 2022-12-06
fix: make sure FunctionWrappers can be pickled

### LINKED ISSUE #200: Instrumentation breaks code that uses the multiprocessing module
Trying to create an AppMap of code that uses the `multiprocessing` module raises an exception.

For example, with a project that looks like this:
```
mp
├── __init__.py
├── hw.py
└── mp.py
```
where `mp/hw.py` contains this
```python
def hw(d):
    d["ret"] = "hello world!"
    print(d["ret"])
```
and `mp/mp.py` contains this
```python
import appmap
from multiprocessing import Manager, Process
from .hw import hw

def do_it():
    with Manager() as manager:
        d = manager.dict()
        p = Process(target=hw, args=(d,))
        p.start()
        p.join()
        return d["ret"]


if __name__ == "__main__":
    r = appmap.Recording()
    with r:
        do_it()
    print(appmap.generation.dump(r))
```
running the command
```shell
$ python -m mp.mp
```
raises an exception that ends like this:
```
  File "/Users/ajp/.asdf/installs/python/3.10.8/lib/python3.10/pickle.py", line 631, in save
    self.save_reduce(obj=obj, *rv)
  File "/Users/ajp/.asdf/installs/python/3.10.8/lib/python3.10/pickle.py", line 729, in save_reduce
    save(func)
  File "/Users/ajp/.asdf/installs/python/3.10.8/lib/python3.10/pickle.py", line 598, in save
    self.save_global(obj)
  File "/Users/ajp/.asdf/installs/python/3.10.8/lib/python3.10/pickle.py", line 1120, in save_global
    __import__(module_name, level=0)
TypeError: __import__() argument 1 must be str, not property
```

(`multiprocessing` is trying to pickle an `appmap.wrapt.FunctionWrapper`. Because of the way `FunctionWrapper` is implemented, `FunctionWrapper.__module__` is a `property` not a `str`, so pickling fails.)

apotterri:
Fixes #200.

The `multiprocessing` modules requires that the callable passed to it be pickleable. Prior to these changes, passing it an instrumented function would fail with the error described in #200.

The fix for this problem is in https://github.com/getappmap/wrapt/commit/7dcec8118747b8a480c1664cad3c16ab9e86f538 . That change updates the `__reduce_ex__` method of `FunctionWrapper` so it just returns the fully-qualified name of the wrapped function. This is enough to enable the function to be pickled and unpickled, as demonstrated by the test in this PR.

You can see the failing behavior in [a previous build](https://app.travis-ci.com/github/getappmap/appmap-python/builds/258568455).

### DIFF
--- a/appmap/test/test_recording.py
+++ b/appmap/test/test_recording.py
@@ -10,6 +10,7 @@ import pytest
 import appmap
 from appmap._implementation.event import Event
 from appmap._implementation.recorder import Recorder, ThreadRecorder
+from appmap.wrapt import FunctionWrapper
 
 from .normalize import normalize_appmap, remove_line_numbers
 
@@ -110,9 +111,22 @@ class TestRecordingWhenEnabled:
 
         rec = appmap.Recording()
         with rec:
+            assert isinstance(modfunc, FunctionWrapper)
             f1 = deepcopy(modfunc)
             f1()
 
+    def test_can_pickle(self):
+        import pickle
+
+        from example_class import (  # pyright: ignore[reportMissingImports] pylint: disable=import-error
+            modfunc,
+        )
+
+        rec = appmap.Recording()
+        with rec:
+            assert isinstance(modfunc, FunctionWrapper)
+            f = pickle.loads(pickle.dumps(modfunc))
+            f()
         evt = rec.events[-2]
         assert evt.event == "call"
         assert evt.method_id == "modfunc"


symwell (approved):
Verified locally it breaks with current wrapt and works with wrapt from 7dcec8.

PR #204, opened 2022-12-08
fix: set appmap_dir in the default config

apotterri:
Make sure appmap_dir is set in the generated appmap.yml, the IDE extensions will perform better when it's there.

### DIFF
--- a/appmap/_implementation/configuration.py
+++ b/appmap/_implementation/configuration.py
@@ -173,7 +173,23 @@ class Config:
 
     @property
     def default(self):
-        return {"name": self.default_name, "packages": self.default_packages}
+        ret = {
+            "name": self.default_name,
+            "language": "python",
+            "packages": self.default_packages,
+        }
+        env = Env.current
+        output_dir = env.output_dir
+        root_dir = env.root_dir
+        try:
+            ret["appmap_dir"] = str(output_dir.relative_to(root_dir))
+        except ValueError:
+            # The only way we can get here is if APPMAP_OUTPUT_DIR is set, and we've already logged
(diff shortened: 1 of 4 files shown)


appmap/_implementation/configuration.py
```diff
@@ -173,7 +173,22 @@ def labels(self):
...
-        return {"name": self.default_name, "packages": self.default_packages}
+        ret = {
+            "name": self.default_name,
+            "packages": self.default_packages,
+        }
+        env = Env.current
+        output_dir = env.output_dir
+        root_dir = env.root_dir
+        try:
+            ret["appmap_dir"] = str(output_dir.relative_to(root_dir))
```
kgilpin, line 184: Please also set `language: python`

appmap/_implementation/configuration.py
```diff
@@ -173,7 +173,23 @@ def labels(self):
...
+        except ValueError:
+            # The only way we can get here is if APPMAP_OUTPUT_DIR is set, and we've already logged
+            # a warning. (Note that PurePath.is_relative_to wasn't added till 3.9, so it can't be
+            # used here.)
+            pass
+
+        return ret
```
symwell, line 192: I think `appmap_dir` and `language` should also be added in `_load_config` in line 229.

apotterri: In discussion with @kgilpin, he made it clear that the output directory shouldn't be configurable, and the value of `appmap_dir` in `appmap.yml` shouldn't have an effect on the agent. There also doesn't seem to be any reason to read `language` from the config file.

So, I'm inclined not to change `_load_config`.

symwell (approved):

PR #211, opened 2023-01-10
Improve error handling of application/json requests

### LINKED ISSUE #209: bad request body causes failures
If `json.loads` can parse the body of an `application/json` request, but it returns something other than a `dict`, both Flask and Django integrations fail to process the request.

For example, a request body of `1` is parseable, but causes the Flask integration raise the exception
```
...
/usr/local/lib/python3.10/site-packages/appmap/_implementation/web_framework.py:164: in before_request_hook
    start, call_event_id = self.before_request_main(rec, request)
/usr/local/lib/python3.10/site-packages/appmap/flask.py:99: in before_request_main
    message_parameters=request_params(request),
/usr/local/lib/python3.10/site-packages/appmap/flask.py:39: in request_params
    params.update(req.json or {})
/usr/local/lib/python3.10/site-packages/werkzeug/datastructures.py:552: in update
    for key, value in iter_multi_items(mapping):
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

mapping = 1

    def iter_multi_items(mapping):
        """Iterates over the items of a mapping yielding keys and values
        without dropping any from more complex structures.
        """
        if isinstance(mapping, MultiDict):
            yield from mapping.items(multi=True)
        elif isinstance(mapping, dict):
            for key, value in mapping.items():
                if isinstance(value, (tuple, list)):
                    for v in value:
                        yield key, v
                else:
                    yield key, value
        else:
>           yield from mapping
E           TypeError: 'int' object is not iterable

/usr/local/lib/python3.10/site-packages/werkzeug/datastructures.py:35: TypeError
```

apotterri:
Fixes #209.

Also, clean up some more issues identified by pylint.

This was originally uncovered by a [test in listenbrainz](https://github.com/metabrainz/listenbrainz-server/blob/7712f5edbcd88ec79444d2bf0fbf606a8f0b2366/listenbrainz/tests/integration/test_api.py#L202).

### DIFF
--- a/_appmap/test/test_flask.py
+++ b/_appmap/test/test_flask.py
@@ -6,22 +6,23 @@ from threading import Thread
 
 import flask
 import pytest
-import requests
 
 from _appmap.env import Env
 from _appmap.metadata import Metadata
-from appmap.flask import AppmapFlask
+from appmap.flask import AppmapFlask, FlaskInserter
 
 from ..test.helpers import DictIncluding
-from .web_framework import (  # pylint: disable=unused-import
-    _TestRecordRequests,
-    exec_cmd,
-    wait_until_port_is,
-)
 
 # Make sure assertions in web_framework get rewritten (e.g. to show
 # diffs in generated appmaps)
-pytest.register_assert_rewrite("test.web_framework")
+pytest.register_assert_rewrite("_appmap.test.web_framework")
+
+# pylint: disable=unused-import,wrong-import-position
+from .web_framework import TestRemoteRecording  # pyright:ignore
+from .web_framework import TestRequestCapture  # pyright: ignore
+from .web_framework import _TestRecordRequests, exec_cmd, wait_until_port_is
+
+# pylint: enable=unused-import
 
 
 @pytest.fixture(name="app")
@@ -29,6 +30,8 @@ def flask_app(data_dir, monkeypatch):
     monkeypatch.syspath_prepend(data_dir / "flask")
 
     Env.current.set("APPMAP_CONFIG", data_dir / "flask" / "appmap.yml")
+
+    monkeypatch.setenv("FLASK_DEBUG", "True")
 
     import app  # pyright: ignore pylint: disable=import-error,import-outside-toplevel
 
--- a/_appmap/web_framework.py
+++ b/_appmap/web_framework.py
@@ -168,7 +168,7 @@ class AppmapMiddleware(ABC):
         return rec, start, call_event_id
 
     @abstractmethod
-    def before_request_main(self, rec):
+    def before_request_main(self, rec, req):
         """Specify the main operations to be performed by a request is processed."""
 
     def after_request_hook(
--- a/_appmap/web_framework.py
+++ b/_appmap/web_framework.py
@@ -11,6 +11,7 @@ import time
 from abc import ABC, abstractmethod
 from contextvars import ContextVar
 from hashlib import sha256
+from json.decoder import JSONDecodeError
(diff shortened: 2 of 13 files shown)


_appmap/test/test_flask.py
```diff
@@ -30,6 +31,8 @@ def flask_app(data_dir, monkeypatch):
 
     Env.current.set("APPMAP_CONFIG", data_dir / "flask" / "appmap.yml")
 
+    monkeypatch.setenv("FLASK_DEBUG", "True")
+
```
symwell, line 35: It seems instead of setting `@pytest.mark.appmap_enabled` in each test, `FLASK_DEBUG` is set in flask_app.

_appmap/web_framework.py
```diff
@@ -233,6 +238,9 @@ def remote_enabled(self):
         """Return True if the AppMap middleware has enabled remote recording, False otherwise."""
 
     def run(self):
+        if not Env.current.enabled:
+            return
+
```
symwell, line 243: This change addresses something different than error handling of application/json requests.  Is there, or should there be, a testcase that fails without this change?

apotterri: Reading this, I realized that my comment on [b24fcd9](https://github.com/getappmap/appmap-python/pull/211/commits/b24fcd9bb1b8df19a39055d233d83ca6dc18ad5e) didn't mention that the tests I restored were broken, and that some of the change address those failures. I'll fix that comment.

Having said that, it's possible that this change wasn't actually necessary,  I'll double-check.

_appmap/test/test_flask.py
apotterri: Yes, this depends on record-by-default working for a Flask app running with `FLASK_DEBUG` set to `True`. Is that what you meant? Do you see a problem here?

symwell: This is what I meant.  I wanted to make sure I understood it.

As long as it's possible to run tests with `FLASK_DEBUG=False` I don't see a problem.

_appmap/web_framework.py
```diff
@@ -168,7 +173,7 @@ def before_request_hook(self, request, request_path):
         return rec, start, call_event_id
 
     @abstractmethod
-    def before_request_main(self, rec):
+    def before_request_main(self, rec, req):
```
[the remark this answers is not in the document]
apotterri: > Should request be passed as a parameter in line 171?
I don't understand what you mean by this question: the call to `before_request_main` at line 171 in the file *does* pass `request`.

There's not a test missing here, pylint identified that signature was incorrect.

_appmap/test/test_flask.py
apotterri: > As long as it's possible to run tests with FLASK_DEBUG=False I don't see a problem.
Well, you can't change this code to say `monkeypatch.setenv("FLASK_DEBUG", "False")`, that would cause tests to fail. If you wanted to set `FLASK_DEBUG=False` in your shell before you ran the tests, though, this would override it, so the tests would pass.

_appmap/web_framework.py
apotterri: Ok, I took this change out and updated the comment.

symwell: Thanks.  Make sure to push the updated branch.

apotterri: Done. (My original attempt to push it failed.)

symwell (approved):

PR #212, opened 2023-01-11
test: add end-to-end smoketest for poetry

### LINKED ISSUE #199: appmap-python has an end-to-end test that uses poetry

apotterri:
Add a quick and dirty end-to-end smoketest that uses poetry.

Fixes #199 .

### DIFF
--- a/ci/test_poetry.sh
+++ b/ci/test_poetry.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -e
+pip install poetry
+
+mkdir /poetry || true
+cd /poetry
+
+poetry init -q
+
+# Yes, we need to set RUNNER, and we need to "poetry run" the script.
+RUNNER="poetry run" poetry run /ci/smoketest.sh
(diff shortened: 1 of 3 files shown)


ci/test_poetry.sh
```diff
@@ -0,0 +1,12 @@
...
+
+set -e
+pip install poetry
+
+mkdir /poetry || true
+cd /poetry
+
+poetry init -q
+
+# Yes, we need to set RUNNER, and we need to "poetry run" the script.
```
dividedmind, line 11: Good call adding this comment :)

dividedmind (changes requested):
It's a start, but perhaps a bit too quick and dirty — I'd like to see the test using `poetry add`, for example; as it is it basically only tests the poetry virtualenv, but not its package dependency machinery (if I understand it correctly). Unless you don't think it's worth it?

apotterri:
I'm not sure it is worth it. With this implementation, it at least shows that installing appmap doesn't break a poetry project, in the same way that test_pipenv.sh shows that it doesn't break a pipenv project.

This test would have prevented us from publising v1.12.4, a release which rendered a poetry project completely non-functional.

dividedmind:
> I'm not sure it is worth it. With this implementation, it at least shows that installing appmap doesn't break a poetry project, in the same way that test_pipenv.sh shows that it doesn't break a pipenv project.
> 
> This test would have prevented us from publising v1.12.4, a release which rendered a poetry project completely non-functional.

Ok, this makes sense. Thanks for explaining the rationale, I'll approve!

dividedmind (approved):

PR #221, opened 2023-02-08
fix: don't instrument extra class members

### LINKED ISSUE #220: appmap-python instruments too many functions
Adding a label preset for a function in a module causes methods on classes in that module to be labeled as well. For example, the presence of the preset `crypto.secure_compare: hmac.compare_digest` (which puts the label `crypto.secure_compare` on the function `hmac.compare_digest`) causes all the methods in the `hmac.HMAC` class to be instrumented as well.

apotterri:
Prior to these changes, labeling a module function by a preset would also cause the members of classes in that module to be labeled. For example, the preset label for hmac.create_digest would cause the member functions of hmac.HMAC to be instrumented as well.

Fixes #220.

### DIFF
--- a/_appmap/importer.py
+++ b/_appmap/importer.py
@@ -21,9 +21,9 @@ Filterable = namedtuple("Filterable", "fqname obj")
 class FilterableMod(Filterable):
     __slots__ = ()
 
-    def __new__(c, mod):
+    def __new__(cls, mod):
         fqname = mod.__name__
-        return super(FilterableMod, c).__new__(c, fqname, mod)
+        return super(FilterableMod, cls).__new__(cls, fqname, mod)
 
     def classify_fn(self, _):
         return FnType.MODULE
@@ -32,9 +32,9 @@ class FilterableMod(Filterable):
 class FilterableCls(Filterable):
     __slots__ = ()
 
-    def __new__(c, cls):
-        fqname = "%s.%s" % (cls.__module__, cls.__qualname__)
-        return super(FilterableCls, c).__new__(c, fqname, cls)
+    def __new__(cls, clazz):
+        fqname = "%s.%s" % (clazz.__module__, clazz.__qualname__)
+        return super(FilterableCls, cls).__new__(cls, fqname, clazz)
 
     def classify_fn(self, static_fn):
         return FnType.classify(static_fn)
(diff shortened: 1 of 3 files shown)


symwell (commented):
Looks good, tests pass.

Noticed something small I'm unsure about.

The currently labeled functions for yaml are:
```
~/src/appmap-python$ cat appmap/labeling/formats.yml | grep yaml
format.yaml.generate:
  - yaml.dump
  - yaml.dump_all
  - yaml.serialize_all
format.yaml.parse:
  - yaml.scan
  - yaml.parse
  - yaml.compose
  - yaml.load
  - yaml.load_all
```

When I add this [diff](https://github.com/getappmap/appmap-python/compare/preset-instrument_20230208...sw/preset-instrument_20230208__debug?expand=1) and run the label testcases, the output includes two functions that aren't in the list of functions from `appmap/labeling/formats.yml`.

The functions are:
```
from_yaml
to_yaml
```

```
$ poetry run pytest -svv _appmap/test/test_labels.py
...
 'yaml': ['dump',
          'dump_all',
          'serialize_all',
          'scan',
          'parse',
          'compose',
          'load',
          'load_all']}
will wrap fn_name scan
will wrap fn_name parse
will wrap fn_name compose
will wrap fn_name load
will wrap fn_name load_all
will wrap fn_name serialize_all
will wrap fn_name dump_all
will wrap fn_name dump
...
will wrap fn_name from_yaml
will wrap fn_name to_yaml
```

apotterri:
I dug in to what's happening with `from_yaml` and `to_yaml`. This uncovered two issues. The `fixup!` commit here fixes one of them. 

The other is described in #224 . That one is going to require some effort to fix, so we need to park it for now.

_appmap/importer.py
```diff
@@ -189,6 +189,9 @@ def instrument_functions(filterable, selected_functions=None):
         fm = FilterableMod(mod)
         if cls.filter_chain.filter(fm):
             instrument_functions(fm, package_functions.get(fm.fqname))
+        elif cls.filter_chain.filter(fm):
+            instrument_functions(fm)
```
symwell, line 193: 1) The two if statements at lines 190 and 192 seem to be the same.  Does `cls.filter_chain.filter()` cause some side effect that would make it return false on the first call and true in the second?  I added debug statements and didn't see the body in the `elif` part to get executed.

What might be happening is that `MatcherFilter` returns either true or false but `NullFilter` always returns false, necessitating the second call to instrument the function.  If so, maybe add a comment in the code explaining part of this.

2) I still see the two functions getting wrapped.  I assume this due to https://github.com/getappmap/appmap-python/issues/224 which we'll park for now.
```
will wrap fn_name from_yaml
will wrap fn_name to_yaml
```

apotterri: Re 1: good catch. This code was supposed to mirror the test at line 202, but I copied and pasted the wrong code. It's fixed now.

Re 2: yes, that's right.

symwell (approved):

PR #240, opened 2023-05-10
Support Werkzeug v2.3, pin to urllib3 v1

### LINKED ISSUE #237: tests fail on master because of updates to Werkzeug and urllib3
One (or more) of the test dependencies have changed, and many tests now fail, as seen in this build: https://app.travis-ci.com/github/getappmap/appmap-python

apotterri:
Updates to some dependencies have broken the build on master. These changes make the build green again, and also fix a couple of pylint warnings.

Fixes #237.

### DIFF
--- a/_appmap/test/test_command.py
+++ b/_appmap/test/test_command.py
@@ -11,8 +11,8 @@ from appmap.command import appmap_agent_init, appmap_agent_status, appmap_agent_
 from .helpers import DictIncluding
 
 
-@pytest.fixture
-def cmd_setup(request, git, data_dir, monkeypatch):
+@pytest.fixture(name="cmd_setup")
+def _cmd_setup(request, git, data_dir, monkeypatch):
     repo_root = git.cwd
     copy_tree(data_dir / request.param, str(repo_root))
     monkeypatch.chdir(repo_root)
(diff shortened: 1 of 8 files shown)


_appmap/test/test_command.py
```diff
@@ -42,40 +43,44 @@ class TestAgentStatus:
...
+                "language": "python",
+                "remoteRecordingCapable": True,
+                "integrationTests": True,
             }
         )
+        assert "agentVersion" in project
+        test_commands = status["test_commands"]
+        assert test_commands[0] == DictIncluding(
+            {"args": [], "framework": "pytest", "command": "pytest"}
+        )
```
dividedmind, line 78: Is there a reason to break this out like that?

dividedmind (approved):

_appmap/test/test_command.py
apotterri: It was originally checking `status` against
```
            {
                "properties": {
                    "config": {"app": "Simple", "present": True, "valid": True},
                    "project": {
                        "agentVersion": "0.0.0",
                        "language": "python",
                        "remoteRecordingCapable": True,
                        "integrationTests": True,
                    },
                },
                "test_commands": [
                    {"args": [], "framework": "pytest", "command": "pytest"}
                ],
                "language": "python",
                "remoteRecordingCapable": True,
                "integrationTests": True,
            }
```
This was failing, because status has a valid value for `agentVersion`, which doesn't match `"0.0.0"`, of course.

When I tried pulling `agentVersion` out of `project`, the comparison still failed, because `status` had it, but the expected value didn't.

I assumed that the `dict` `>=` comparison just wasn't doing the right thing with nested `dict`s, so I unnested them.

dividedmind: Ah, ok, makes sense. That's fine, thanks. (Though I wonder if `"project": DictIncluding(...)` would work instead?)

PR #243, opened 2023-05-22
fix: Don't record requests in test cases

### LINKED ISSUE #234: Request recordings should not be generated by tests
Analogous to the way appmap-ruby works, there's no need to generate request recordings when running tests. They're a subset of the events in the test recordings, which isn't useful, and will confuse users.

dividedmind:
Fixes #234

### DIFF
--- a/_appmap/env.py
+++ b/_appmap/env.py
@@ -1,5 +1,6 @@
 """Initialize from the environment"""
 
+from contextlib import contextmanager
 import logging
 import logging.config
 import os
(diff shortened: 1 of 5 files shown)


kgilpin:
In case it's helpful, this is how we test the interaction between these different environment variables and other factors in appmap-ruby - https://github.com/getappmap/appmap-ruby/blob/master/spec/detect_enabled_spec.rb

apotterri (approved):
Looks good, thanks for doing this.

PR #250, opened 2023-07-08
feat: Report test_failure when a test fails

### LINKED ISSUE #248: Report test_failure when a test fails
See appmap spec for details on this metadata field.

dividedmind:
Fixes #248

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,10 @@ python:
 # https://github.com/travis-ci/travis-ci/issues/1147#issuecomment-441393807
 if: type != push OR branch = master OR branch =~ /^v\d+\.\d+(\.\d+)?(-\S*)?$/
 
-before_install: pip install --upgrade pip 'setuptools==65.6.2' 'poetry>=1.2.0'
+before_install: |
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -11,7 +11,10 @@ python:
 # https://github.com/travis-ci/travis-ci/issues/1147#issuecomment-441393807
 if: type != push OR branch = master OR branch =~ /^v\d+\.\d+(\.\d+)?(-\S*)?$/
 
-before_install: pip install --upgrade pip 'setuptools==65.6.2' 'poetry>=1.2.0'
+before_install: |
+  curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain stable --profile minimal
```
apotterri, line 15: What changed about 3.10 that caused it to require rust?

apotterri (approved):
LGTM, thanks.

.travis.yml
apotterri: But that didn't cause the build to fail? Weird.

dividedmind: It did, that's why I had to fix it :) I think there might have been a recent change in one of the dependencies. What's weirder for me is that it only seems to be required for 3.10 o_O

PR #266, opened 2023-11-04
ci, wip: remove rust

apotterri:


apotterri:
Rust isn't the thing that's causing this to take 5+min to install on 3.10, rapidfuzz is. Looks like something Travis needs to sort out....

PR #271, opened 2024-01-22
Exclude the appmap test files

### LINKED ISSUE #272: `appmap.yml` files are not shipped in releases

petecheslock:
Taking a first pass of this - see the issue here: https://github.com/getappmap/appmap-js/issues/1555. AppMap spawns separate index files for every appmap.yaml - so excluding the test data from the package.  

Fixes #272


pyproject.toml
```diff
@@ -22,7 +22,10 @@ classifiers = [
         'Topic :: Software Development :: Documentation'
 ]
 include = ['appmap.pth']
-exclude = ['_appmap/wrapt']
+exclude = [
+    '_appmap/wrapt',
+    '_appmap/test/data/*/appmap.yml'
```
kgilpin, line 27: Maybe all of `_appmap/test` would be fine?

apotterri:
It's important to ensure that all the files necessary to run the tests are included in the source distribution (sdist) that gets created by poetry (and uploaded to the PyPI).

If they are, then this change is fine, and in fact can be improved by Kevin's suggestion of excluding all of `_appmap/test`.

If they're not included, though, you'll need to dig into the poetry doc and see if there's a different approach.

apotterri:
Also, the Travis build is red. Looks like you may need to pin pytest-django to an older version.

kgilpin:
> It's important to ensure that all the files necessary to run the tests are included in the source distribution (sdist) that gets created by poetry (and uploaded to the PyPI).
> 
> If they are, then this change is fine, and in fact can be improved by Kevin's suggestion of excluding all of `_appmap/test`.
> 
> If they're not included, though, you'll need to dig into the poetry doc and see if there's a different approach.

So there are different types of distributions? The package distribution and the source distribution?

petecheslock:
Thanks @apotterri i'll check out some docs to see what some ideas could be to solve this.  I suppose if there isn't an easy way to exclude those files then the alternative solution would be for me to just modify the tests to use an appmap.yml with a different name and just change those filenames accordingly?

petecheslock:
Yea i think that is correct.  I found this in the poetry project which gives an example of how i might exclude those files from the wheel distribution while also keeping them included in the source.  https://github.com/python-poetry/poetry/issues/3380#issuecomment-734915524

going to give this a try and update this PR when i have it working.

petecheslock:
Ok - i think i have this working.  With the recent change i do the following

`poetry build`

From there i get this:

```
Building appmap (1.18.1)
  - Building sdist
  - Built appmap-1.18.1.tar.gz
  - Building wheel
  - Built appmap-1.18.1-py3-none-any.whl
```

if i unzip the source tar.gz the `_appmap/test` directory exits

When i unzip the wheel - the test directory does NOT exist.

Then in my django oscar project i install the wheel from the local built version

```
$ head requirements.txt                                                                                                                                                                   
file:////Users/petecheslock/repos/appmap-python/dist/appmap-1.18.1-py3-none-any.whl
```

Then after install i checked the log and only one indexer is running (which is what is expected).

apotterri:
> if i unzip the source tar.gz the _appmap/test directory exits

> When i unzip the wheel - the test directory does NOT exist.

Yup, this is the desired behavior. 👍

petecheslock:
Thanks @apotterri  - i updated the files (i thiink correctly) to use a working version on pytest-django and the tests now pass.  Can you let me know if that is correct and then this is ready for review.

apotterri (approved):
LGTM. Thanks for doing this.

PR #273, opened 2024-01-22
fix: Exclude the appmap test files (#271)

petecheslock:
* Exclude the appmap test files

* Update glob to exclude more files

* Include the necessary files only for the sdist exclude them otherwise

* Use a lower working version of pytest-django

* Update the tox config correctly


petecheslock:
@apotterri i think this might work?

PR #274, opened 2024-02-05
Support python 3.12

### LINKED ISSUE #253: Python 3.12 is tested with our Python agent
Using this app: https://github.com/jainamoswal/Flask-Example

We need to verify that 3.12 (when it is officially released) will work with our Python agent. 3.12rc2 is available now: https://docs.python.org/3.12/whatsnew/3.12.html

apotterri:
These changes ensure that the agent works with 3.12.

Fixes #253 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ os: linux
 dist: jammy
 language: python
 python:
+- "3.12"
 - "3.11"
 - "3.10"
 - "3.9.14"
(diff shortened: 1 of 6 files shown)


apotterri:
Hmmm, I noticed there may be a problem when shutting down after the tests run. That probably needs to be investigated before this gets merged.

apotterri:
Looks like the shutdown issue was related to some local changes I had in my repo. I don't see any issue with this PR, should be good to go.

dustinbyrne (approved):

PR #285, opened 2024-02-27
Make sure an Exception in a Flask route gets captured, test Flask 3

### LINKED ISSUE #286: exceptions in a Flask route handler don't get captured
For example, when this code gets executed:

```python
@app.route("/exception")
def raise_exception():
    raise Exception("An exception")
```

an `http_server_request` will be added to the AppMap, but without a corresponding `http_server_response`.

apotterri:
Also, clean up some references to python 3.7 that got missed in the last PR, add a TRACE log level, and address a few more lint warnings.

Fixes #286 .


_appmap/env.py
```diff
@@ -1,11 +1,12 @@
 """Initialize from the environment"""
 
-from contextlib import contextmanager
 import logging
 import logging.config
 import os
+from contextlib import contextmanager
 from os import environ
 from pathlib import Path
+from typing import cast
```
dividedmind, line 9: Seems like this is unused in this commit?

_appmap/web_framework.py
```diff
@@ -109,7 +115,7 @@ def create_appmap_file(
         + " "
         + request_path_info
         + " ("
-        + str(response.status_code)
+        + str(status)
         + ") - "
         + start_time.strftime("%T.%f")[:-3]
```
dividedmind, line 120: Interesting that pylint doesn't suggest using an f-string here.

tox.ini
```diff
@@ -13,7 +13,8 @@ deps=
     poetry
     pytest-django <4.8
     web: Django >=4.0, <5.0
-    web: Flask >= 2, <3
+    web: Flask >2.0
```
dividedmind, line 16: Should it say `Flask >= 3.0` for clarity?

dividedmind (approved):

_appmap/web_framework.py
apotterri: `consider-using-f-string` is currently disabled. Turning it on doesn't generate that many suggestions, so I'll fix them in a future PR.

Even with it turned on, pylint doesn't see that this could be an f-string. Maybe it's a little too complicated....

Issue #256, opened 2023-10-13
Simple media GET requests should not be saved as request recordings


brikelly:

virajkanwade:
@apotterri i wouldn't call this a bug. More like a feature enhancement where it's a config flag to enable/disable the media requests tracking.
Somebody might actually want to track media requests.

PR #288, opened 2024-02-29
[#256] Flag to disable static request recording 

virajkanwade:
`APPMAP_ENABLED_STATIC_REQUEST_RECORDINGS=false flask --app main run`

Submitting draft code - please adapt as seen fit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+# [1.20.0](https://github.com/getappmap/appmap-python/compare/v1.19.1...v1.20.0) (2024-03-15)
+
+
+### Bug Fixes
+
+* log to a file by default ([3da004f](https://github.com/getappmap/appmap-python/commit/3da004f4ba67b7e3a923fc76ef733eec51a496c7))
+
+
+### Features
+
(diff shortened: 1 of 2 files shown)


virajkanwade:
@apotterri @brikelly Let me know your thoughts.

_appmap/env.py
```diff
@@ -46,6 +46,7 @@ def __init__(self, env=None, cwd=None):
 
         self._configure_logging()
         self._enabled = self._env.get("APPMAP", "").lower() != "false"
+        self._enabled_static_request_recordings = self._env.get("APPMAP_ENABLED_STATIC_REQUEST_RECORDINGS", "").lower() != "false"
```
apotterri, line 49: Maybe `APPMAP_RECORD_STATIC_REQUESTS` instead? That would be somewhat more consistent with other env vars.

apotterri (changes requested):
Thanks very much for this. I suggested one minor change.

It was also be very helpful it you could add a test or two.

(Also, FYI, Brian is no longer with AppMap.)

_appmap/env.py
virajkanwade: @apotterri I had thought of it at first, but then thought it might interfere with https://github.com/getappmap/appmap-python/blob/master/_appmap/env.py#L18

PR #293, opened 2024-03-16
add: ruff linter

mavwolverine:
`poetry run ruff check --fix appmap`

https://docs.astral.sh/ruff/

### DIFF
--- a/appmap/unittest.py
+++ b/appmap/unittest.py
@@ -4,4 +4,4 @@ logger = Env.current.getLogger(__name__)
 
 if not Env.current.is_appmap_repo and Env.current.enables("unittest"):
     logger.debug("Test recording is enabled (unittest)")
-    import _appmap.unittest  # pyright: ignore pylint: disable=unused-import
+    import _appmap.unittest  # pyright: ignore pylint: disable=unused-import # noqa: F401
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,6 +79,7 @@ uvicorn = "^0.27.1"
 fastapi = "^0.110.0"
 httpx = "^0.27.0"
 pytest-env = "^1.1.3"
+ruff = "^0.3.3"
 
 [build-system]
 requires = ["poetry-core>=1.1.0"]
--- a/appmap/__init__.py
+++ b/appmap/__init__.py
@@ -4,7 +4,7 @@ from _appmap import generation  # noqa: F401
 from _appmap.env import Env  # noqa: F401
 from _appmap.importer import instrument_module  # noqa: F401
 from _appmap.labels import labels  # noqa: F401
-from _appmap.noappmap import decorator as noappmap
+from _appmap.noappmap import decorator as noappmap  # noqa: F401
 from _appmap.recording import Recording  # noqa: F401
 
 try:
--- a/appmap/http.py
+++ b/appmap/http.py
@@ -52,7 +52,7 @@ class HTTPConnectionPatch:
         if not hasattr(request, "headers"):
             request["headers"] = {}
         headers = request["headers"]
-        if not header in headers:
+        if header not in headers:
             headers[header] = []
         headers[header].extend(values)
         orig(self, header, *values)
--- a/appmap/labeling/__init__.py
+++ b/appmap/labeling/__init__.py
@@ -8,7 +8,7 @@ from functools import lru_cache
 import yaml
 from importlib_resources import files
 
-from _appmap.labels import LabelSet
+from _appmap.labels import LabelSet  # noqa: F401
 
 
 @lru_cache(maxsize=None)
--- a/appmap/pytest.py
+++ b/appmap/pytest.py
@@ -80,5 +80,5 @@ if not Env.current.is_appmap_repo and Env.current.enables("pytest"):
             try:
                 with testing_framework.collect_result_metadata(metadata):
                     result.get_result()
-            except:  # pylint: disable=bare-except
+            except:  # pylint: disable=bare-except # noqa: E722
                 pass  # exception got recorded in metadata


appmap/unittest.py
```diff
@@ -4,4 +4,3 @@
 
 if not Env.current.is_appmap_repo and Env.current.enables("unittest"):
     logger.debug("Test recording is enabled (unittest)")
-    import _appmap.unittest  # pyright: ignore pylint: disable=unused-import
```
apotterri, line 7: This import needs to be retained, so it should be marked with `# noqa: F401` rather than deleted.

pyproject.toml
```diff
@@ -49,6 +49,7 @@ packaging = ">=19.0"
 #
 # So, if you'd like to run the tests outside of tox, run `pip install -r requirements-dev.txt` to
 # install it and the rest of the dev dependencies.
+ruff = "^0.3.3"
```
apotterri, line 52: This should be in the `dev` dependencies (below).

apotterri:
We're already using `pylint` for linting: https://github.com/getappmap/appmap-python/blob/master/tox.ini#L24 . I'm not opposed to switching to `ruff`, but if we do, we should remove `pylint`. This should probably also include a config file for `ruff` that more or less matches the `pylint` config.

apotterri:
We're using [semantic-release](https://semantic-release.gitbook.io/semantic-release/) for `appmap-python`, so the subject lines for commits need to follow this format: https://github.com/angular/angular/blob/main/CONTRIBUTING.md#-commit-message-format .

(Our own `CONTRIBUTING.md` really needs an update that mentions this....)

apotterri (changes requested):
Thanks very much for this.

appmap/unittest.py
virajkanwade: Fixed

virajkanwade:
> We're using [semantic-release](https://semantic-release.gitbook.io/semantic-release/) for `appmap-python`, so the subject lines for commits need to follow this format: https://github.com/angular/angular/blob/main/CONTRIBUTING.md#-commit-message-format .
> 
> (Our own `CONTRIBUTING.md` really needs an update that mentions this....)

Updated the commit message. Hopefully chose the correct one.

virajkanwade:
> We're already using `pylint` for linting: https://github.com/getappmap/appmap-python/blob/master/tox.ini#L24 . I'm not opposed to switching to `ruff`, but if we do, we should remove `pylint`. This should probably also include a config file for `ruff` that more or less matches the `pylint` config.

Most of projects are now moving from pylint to ruff.
https://docs.astral.sh/ruff/faq/#how-does-ruffs-linter-compare-to-pylint

Also, the fastapi unused wsgimiddleware import, ruff highlighted it, but I didn't notice it as such in the pylint output.

This was a suggestion for replacing pylint with ruff and not both together.

I am not aware of other places where you might have to replace pylint in the dev environments/ CI etc.

apotterri:
> > We're already using `pylint` for linting: https://github.com/getappmap/appmap-python/blob/master/tox.ini#L24 . I'm not opposed to switching to `ruff`, but if we do, we should remove `pylint`. This should probably also include a config file for `ruff` that more or less matches the `pylint` config.
> 
> Most of projects are now moving from pylint to ruff. https://docs.astral.sh/ruff/faq/#how-does-ruffs-linter-compare-to-pylint
Yup, seems like a good idea to me.

> 
> Also, the fastapi unused wsgimiddleware import, ruff highlighted it, but I didn't notice it as such in the pylint output.
> 
> This was a suggestion for replacing pylint with ruff and not both together.
Understood. I wasn't suggesting this either.

> 
> I am not aware of other places where you might have to replace pylint in the dev environments/ CI etc.
The main thing we're using `pylint` to try to keep the code from getting worse: https://github.com/getappmap/appmap-python/blob/master/tox.ini#L24 .

We need to retain this functionality if we switch to ruff, so this PR would need to include changes to do that, as well.

virajkanwade:
@apotterri Just to confirm, this is executed when running tox? Also, what is the expected outcome here? Tests should fail if there is any pylint error? As pylint-exit only returns 1 on `Fatal message issued` or `Usage error`

apotterri:
Yes, it's run when the `tox` environment is `py310-web`: https://app.travis-ci.com/github/getappmap/appmap-python/jobs/619305807#L579 .

pylint computes a score for the code, and has a setting that controls the minimum score: https://github.com/getappmap/appmap-python/blob/master/pylintrc#L3 . If the code that's getting built has a score that's below that threshold, pylint-exit returns non-zero.

virajkanwade:
> Yes, it's run when the `tox` environment is `py310-web`: https://app.travis-ci.com/github/getappmap/appmap-python/jobs/619305807#L579 .
> 
> pylint computes a score for the code, and has a setting that controls the minimum score: https://github.com/getappmap/appmap-python/blob/master/pylintrc#L3 . If the code that's getting built has a score that's below that threshold, pylint-exit returns non-zero.

Doesn't seem to be working that way.

```
# Specify a score threshold under which the program will exit with error.
fail-under=9.83
```

```
------------------------------------------------------------------
Your code has been rated at 9.80/10 (previous run: 9.80/10, +0.00)

The following messages were raised:

  - error message issued
  - warning message issued
  - refactor message issued
  - convention message issued

No fatal messages detected.  Exiting gracefully...
```

apotterri:
That would be bug, then. It's intended to work that way, and I thought that was the description from pylint-exit.

apotterri:
> That would be bug, then. It's intended to work that way, and I thought that was the description from pylint-exit.

Looking again at the doc for pylint-exit, I see that I misread it. It doesn't clean to exit non-zero if the `fail_under` check fails. So, there's definitely a bug in the build that needs to be fixed.

lightningRalf:
So just to understand how to move this along is by adding mypy as a separate PR that does the type checking? And if this is implemented in a separate PR than this could move forward?

lightningRalf:
> So just to understand how to move this along is by adding mypy as a separate PR that does the type checking? And if this is implemented in a separate PR than this could move forward?

So now we have ty in a nice place, i would actually use that over mypy.

PR #298, opened 2024-04-15
feat: don't record by default, add runner

### LINKED ISSUE #294: Disable the agent by default
The VSCode extension seems to run Python commands which trigger log files to fill up the root of the project directory.

We either need to disable logging by default, or disable it under this specific circumstance.

From a user:
> It keeps creating appmap.yml and pollutes my working directory with myriad of appmap.logs.

apotterri:
**https://github.com/getappmap/applandinc.github.io/pull/1325 needs to be merged before this.**

These change begin the transition of a change in the agent's behavior. In version 2.0 (the next version released), instrumentation will no longer be automatically enabled after it's installed in a Python environment. To create recordings, the environment variable `APPMAP` must be set to `true` (case-insensitive).

In addition, there is now a runner script, `appmap-python`. It can be used to control the environment variables, including `APPMAP_DISABLE_LOG_FILE`.

Fixes #294 .

### DIFF
--- a/_appmap/env.py
+++ b/_appmap/env.py
@@ -3,6 +3,7 @@
 import logging
 import logging.config
 import os
+import warnings
 from contextlib import contextmanager
 from datetime import datetime
 from os import environ
@@ -10,6 +11,20 @@ from pathlib import Path
 from typing import cast
 
 from . import trace_logger
+
+_ENABLED_BY_DEFAULT_MSG = """
+
+The APPMAP environment variable is unset. Your code will be
+instrumented and recorded according to the configuration in appmap.yml.
+
+Starting with version 2, this behavior will change: when APPMAP is
+unset, no code will be instrumented. You will need to use the
+appmap-python script to run your application, or explicitly set
+APPMAP.
+
+Visit https://appmap.io/docs/reference/appmap-python.html#appmap-python-script for more
+details.
+"""
 
 _cwd = Path.cwd()
 _bootenv = environ.copy()
@@ -37,16 +52,21 @@ class _EnvMeta(type):
 
 class Env(metaclass=_EnvMeta):
     def __init__(self, env=None, cwd=None):
+        warnings.filterwarnings("once", _ENABLED_BY_DEFAULT_MSG)
+
         # root_dir and root_dir_len are going to be used when
         # instrumenting every function, so preprocess them as
         # much as possible.
+
         self._cwd = cwd or _cwd
         self._env = _bootenv.copy()
         if env:
             self._env.update(env)
 
         self._configure_logging()
-        self._enabled = self._env.get("APPMAP", "").lower() != "false"
+        enabled = self._env.get("_APPMAP", None)
+        self._enabled_by_default = enabled is None
+        self._enabled = enabled is None or enabled.lower() != "false"
 
         self._root_dir = str(self._cwd) + "/"
         self._root_dir_len = len(self._root_dir)
@@ -81,6 +101,14 @@ class Env(metaclass=_EnvMeta):
     @property
     def output_dir(self):
         return self._output_dir
+
+    @property
+    def enabled_by_default(self):
+        return self._enabled_by_default
+
+    def warn_enabled_by_default(self):
+        if self._enabled_by_default:
+            warnings.warn(_ENABLED_BY_DEFAULT_MSG, category=DeprecationWarning, stacklevel=2)
 
     @property
     def enabled(self):
--- a/appmap/__init__.py
+++ b/appmap/__init__.py
@@ -1,38 +1,44 @@
 """AppMap recorder for Python"""
+import os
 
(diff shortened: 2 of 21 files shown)


kgilpin (changes requested):
See comments on the related issue. We can make changes to the agent launch capability but not incompatible ones without a robust migration.

appmap/__init__.py
```diff
@@ -1,38 +1,43 @@
...
-
-def enabled():
-    return Env.current.enabled
+import os
+
+# Note that we need to guard these imports with a conditional, rather than
+# putting them in a function and conditionally calling the function. If we
+# execute the imports in a function, the modules all get put into the funtion's
+# globals, rather than into appmap's globals.
+if os.environ.get("APPMAP", "false").upper() == "TRUE":
```
kgilpin, line 8: Due to the drastic nature of this change, this would be a major version revision. 

The library on the current major version should warn loudly that this change is going to happen. 

We will also need to notify Python users as robustly as we can, eg with an email.

apotterri: > Due to the drastic nature of this change, this would be a major version revision.
Ok. I'll mark this a breaking change.

> 
> The library on the current major version should warn loudly that this change is going to happen.
I'll open separate PR for this that we'll merge before this one.

> 
> We will also need to notify Python users as robustly as we can, eg with an email.
I'm not sure what the mechanism for sending such an email would be, so I'll need some help with this.

apotterri: I've added a warning that gets displayed when the user is relying on the default behavior.

For example, when running `pytest`, they'll see this:
```
======================== 1 xfailed, 1 warning in 1.46s =========================
pytest-xprocess reminder::Be sure to terminate the started process by running 'pytest --xkill' if you have not explicitly done so in your fixture with 'xprocess.getinfo(<process_name>).terminate()'.
/Users/ajp/src/applandinc/appmap-python/appmap/pytest.py:30: DeprecationWarning: 

The APPMAP environment variable is unset. Your code will be
instrumented and recorded according to the configuration in appmap.yml.

Starting with version 2, this behavior will change: when APPMAP is
unset, no code will be instrumented. You will need to use the
appmap-python script to run your application, or explicitly set
APPMAP.

Visit https://appmap.io/docs/reference/appmap-python.html#appmap-python-script for more
details.

  Env.current.warn_enabled_by_default()
[2024-04-23 09:06:30,521] WARNING _appmap.metadata: Failed running 'git rev-parse HEAD', fatal: not a git repository (or any of the parent directories): .git

PASSED
```
This is the standard way that`pytest` reports warnings that occur during tests.

When running a Django app, they'll see this:
```
/Users/ajp/src/applandinc/appmap-python/.direnv/python-3.10.13/bin/python manage.py runserver 127.0.0.1:8000
[2024-04-23 09:09:30,794] WARNING _appmap.env: Setting APPMAP_OUTPUT_DIR is not supported
[2024-04-23 09:09:30,854] WARNING _appmap.env: Setting APPMAP_OUTPUT_DIR is not supported
[2024-04-23 08:09:32,515] WARNING _appmap.env: Setting APPMAP_OUTPUT_DIR is not supported
[2024-04-23 08:09:32,551] WARNING _appmap.env: Setting APPMAP_OUTPUT_DIR is not supported
Performing system checks...

System check identified no issues (0 silenced).
/Users/ajp/src/applandinc/appmap-python/_appmap/web_framework.py:250: DeprecationWarning: 

The APPMAP environment variable is unset. Your code will be
instrumented and recorded according to the configuration in appmap.yml.

Starting with version 2, this behavior will change: when APPMAP is
unset, no code will be instrumented. You will need to use the
appmap-python script to run your application, or explicitly set
APPMAP.

Visit https://appmap.io/docs/reference/appmap-python.html#appmap-python-script for more
details.

  Env.current.warn_enabled_by_default()
April 23, 2024 - 08:09:33
Django version 5.0.4, using settings 'djangoapp.settings'
Starting development server at http://127.0.0.1:8000/
Quit the server with CONTROL-C.
```
Again, standard reporting.

The user will also see a warning when doing process recording, as well as code-block recording.

You asked for the help text to show an example command line, but the way the Standard Library's `warnings` module works, as well as the way `argv` is handled, didn't seem to allow this. I'm hoping that linking to the doc is an acceptable compromise.

_appmap/env.py
```diff
@@ -11,6 +12,20 @@
...
 from . import trace_logger
 
+_ENABLED_BY_DEFAULT_MSG = """
+
+The APPMAP environment variable is unset. Your code will be
+instrumented and recorded according to the configuration in appmap.yml.
+
+Starting with version 2, this behavior will change: when APPMAP is
+unset, no code will be instrumented. You will need to use the
+appmap-python script to run your application, or explicitly set
```
kgilpin, line 22: Does this message show if they are already running the launcher script or have set APPMAP?

kgilpin (approved):

PR #303, opened 2024-05-12
feat: Check malformed path entries

### LINKED ISSUE #302: the agent should give an error for malformed `path` entries
The agent should check the format of a `path` entry in `appmap.yml`, and fail if it finds a bad one. In particular, it should detect entries that contain path separators (`/`, `\`) and offer a hint that `path` entries are module names, not directory paths.

zermelo-wisen:
Fixes #302

Warning and dropping malformed configuration file package path entries, which cannot be parsed as Python module names: `mod1/mod2`, `mod1\mod2`, `123`, `.`, ` `.

### DIFF
--- a/_appmap/configuration.py
+++ b/_appmap/configuration.py
@@ -2,6 +2,7 @@
 Manage Configuration AppMap recorder for Python.
 """
 
+import ast
 import importlib.metadata
 import inspect
 import os
@@ -125,6 +126,8 @@ def find_top_packages(rootdir):
 
     return packages
 
+class AppMapInvalidConfigException(Exception):
+    pass
 
 class Config(metaclass=SingletonMeta):
     """Singleton Config class"""
@@ -179,7 +182,7 @@ class Config(metaclass=SingletonMeta):
         root_dir = Env.current.root_dir
         return [{"path": p} for p in find_top_packages(root_dir)]
 
-    def _load_config(self):
+    def _load_config(self, show_warnings=False):
         self._config = {"name": None, "packages": []}
 
         # Only use a default config if the user hasn't specified a
@@ -192,7 +195,7 @@ class Config(metaclass=SingletonMeta):
 
         env = Env.current
         config_dir = env.root_dir
-        
+
         path = _resolve_relative_to(Path(env_config_filename), Path(config_dir))
         if not path.is_file():
             # search config file in parent directories up to
@@ -221,6 +224,8 @@ class Config(metaclass=SingletonMeta):
                         self._config["name"] = self.default_name
                     if "packages" not in self._config:
                         self._config["packages"] = self.default_packages
+                    else:
+                        self._drop_malformed_package_paths(show_warnings)
 
                     # Is appmap_dir specified?
                     appmap_dir = (
@@ -283,6 +288,43 @@ It will be created with this configuration:
             modules[mod] = modules[mod] + [name] if mod in modules else [name]
 
         self.package_functions.update(modules)
+
+    def _drop_malformed_package_paths(self, show_warnings):
+        invalid_items = []
+        for item in self._config["packages"]:
+            # it can be a "dist" entry
+            if "path" not in item:
+                continue
+
+            path = item.get("path")
+            if path is None:
+                if show_warnings:
+                    logger.warning("Missing path value in configuration file.")
+                invalid_items.append(item)
+                continue
+
+            if not self._check_path_value(path):
+                has_separator = isinstance(path, str) and ('/' in path or '\\' in path)
+                if show_warnings:
+                    logger.warning(
+                        f"Malformed path value '{path}' in configuration file. "
+                        "Path entries must be module names"
+                        f"{' not directory paths' if has_separator else ''}.",
+                        stack_info=False,
+                    )
+                invalid_items.append(item)
+                continue
+
+        if len(invalid_items) > 0:
+            self._config["packages"] = [item for item in self._config["packages"]
+                                        if item not in invalid_items]
+
+    def _check_path_value(self, value):
+        try:
+            ast.parse(f"import {value}")
+            return True
+        except SyntaxError:
+            return False
 
 
 def startswith(prefix, sequence):
@@ -430,6 +472,7 @@ def initialize():
 initialize()
 
 c = Config.current
+c._load_config(show_warnings=True)
 logger.info("config: %s", c._config)
 logger.debug("package_functions: %s", c.package_functions)
 logger.info("env: %r", os.environ)
--- a/_appmap/configuration.py
+++ b/_appmap/configuration.py
@@ -14,6 +14,7 @@ import yaml
 from yaml.parser import ParserError
 
 from _appmap.labels import LabelSet
+from _appmap.singleton import SingletonMeta
 from appmap.labeling import presets as label_presets
 
 from . import utils
@@ -125,24 +126,10 @@ def find_top_packages(rootdir):
     return packages
 
 
-class Config:
+class Config(metaclass=SingletonMeta):
     """Singleton Config class"""
 
-    _instance = None
-
-    def __new__(cls):
-        if cls._instance is None:
-            logger.trace("Creating the Config object")
-            cls._instance = super(Config, cls).__new__(cls)
-
-            cls._instance._initialized = False
-
-        return cls._instance
-
     def __init__(self):
-        if self._initialized:
-            return
-
         self.file_present = False
         self.file_valid = False
         self.package_functions = {}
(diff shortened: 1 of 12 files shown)


_appmap/configuration.py
```diff
@@ -116,6 +117,8 @@ def excluded(d):
 
     return packages
 
+class AppMapPackagePathException(Exception):
```
apotterri, line 120:
```suggestion
class AppMapInvalidConfigException(Exception):
```

_appmap/configuration.py
```diff
@@ -219,7 +222,8 @@ def _load_config(self):
                         self._config["name"] = self.default_name
                     if "packages" not in self._config:
                         self._config["packages"] = self.default_packages
-
+                    else:
+                        self._check_path_values()
```
apotterri, line 226: Having `_check_path_values` raise an exception is good, because it makes it easy to test.

However, showing the user a stack trace here isn't going to be helpful. Instead, this should catch the exception, show the user the message, and call `sys.exit(1)`.

_appmap/configuration.py
```diff
@@ -269,6 +273,24 @@ def _load_functions(self):
...
         self.package_functions.update(modules)
 
+    def _check_path_values(self):
+        for item in self._config["packages"]:
+            path: str = item.get("path")
+            if path and not self._check_path_value(path):
+                has_separator = '/' in path or '\\' in path
+                raise AppMapPackagePathException(
+                    f"Malformed path value '{path}' in configuration file. "
+                    "Path entries should be module names"
```
apotterri, line 283:
```suggestion
                    "Path entries must be module names"
```

_appmap/configuration.py
```diff
@@ -269,6 +273,24 @@ def _load_functions(self):
 
         self.package_functions.update(modules)
 
+    def _check_path_values(self):
+        for item in self._config["packages"]:
+            path: str = item.get("path")
+            if path and not self._check_path_value(path):
```
apotterri, line 279: This code should also raise an exception if `path` is `None`. This case is currently not handled well, and produces an ugly error message.

apotterri (changes requested):
Thanks for this. Just a couple of minor changes to make, then it'll be good to go.

_appmap/configuration.py
apotterri: What's the value in continuing in this case? The user has made it clear that they'd like to get recordings (by updating their `appmap.yml`), but the change they've made makes that impossible. If it was me, I'd like to learn that immediately, rather than having the message scroll off my screen as all my tests run....

kgilpin: OK

apotterri:
I gave these changes a try. Because of the way `appmap` gets imported (using a `.pth` file), the UX is terrible:
```
[2024-05-14 06:25:39,785] ERROR _appmap.configuration: Malformed path value '.' in configuration file. Path entries must be module names.
Fatal Python error: init_import_site: Failed to import the site module
Python runtime state: initialized
Traceback (most recent call last):
  File "/Users/ajp/src/applandinc/worktrees/appmap-python/_appmap/configuration.py", line 227, in _load_config
    self._check_path_values()
  File "/Users/ajp/src/applandinc/worktrees/appmap-python/_appmap/configuration.py", line 294, in _check_path_values
    raise AppMapInvalidConfigException(
_appmap.configuration.AppMapInvalidConfigException: Malformed path value '.' in configuration file. Path entries must be module names.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/ajp/.asdf/installs/python/3.10.13/lib/python3.10/site.py", line 617, in <module>
    main()
  File "/Users/ajp/.asdf/installs/python/3.10.13/lib/python3.10/site.py", line 610, in main
    execsitecustomize()
  File "/Users/ajp/.asdf/installs/python/3.10.13/lib/python3.10/site.py", line 549, in execsitecustomize
    import sitecustomize
  File "/Users/ajp/src/applandinc/worktrees/appmap-python/_appmap/test/data/fastapi/init/sitecustomize.py", line 1, in <module>
    import appmap
  File "/Users/ajp/src/applandinc/worktrees/appmap-python/appmap/__init__.py", line 9, in <module>
    from _appmap import generation  # noqa: F401
  File "/Users/ajp/src/applandinc/worktrees/appmap-python/_appmap/__init__.py", line 1, in <module>
    from . import configuration
  File "/Users/ajp/src/applandinc/worktrees/appmap-python/_appmap/configuration.py", line 450, in <module>
    initialize()
  File "/Users/ajp/src/applandinc/worktrees/appmap-python/_appmap/configuration.py", line 445, in initialize
    Config().initialize()
  File "/Users/ajp/src/applandinc/worktrees/appmap-python/_appmap/configuration.py", line 146, in __init__
    self._load_config()
  File "/Users/ajp/src/applandinc/worktrees/appmap-python/_appmap/configuration.py", line 230, in _load_config
    sys.exit(1)
SystemExit: 1
```

So, we need to change this. When an invalid `path` is detected`, we should just log a warning and ignore it. This needs to happen without raising an exception, and the agent should not exit.

apotterri (changes requested):
https://github.com/getappmap/appmap-python/pull/303#issuecomment-2109843178

zermelo-wisen:
> I gave these changes a try. Because of the way `appmap` gets imported (using a `.pth` file), the UX is terrible:
> 
> ```
> [2024-05-14 06:25:39,785] ERROR _appmap.configuration: Malformed path value '.' in configuration file. Path entries must be module names.
> Fatal Python error: init_import_site: Failed to import the site module
> Python runtime state: initialized
> Traceback (most recent call last):
>   File "/Users/ajp/src/applandinc/worktrees/appmap-python/_appmap/configuration.py", line 227, in _load_config
>     self._check_path_values()
>   File "/Users/ajp/src/applandinc/worktrees/appmap-python/_appmap/configuration.py", line 294, in _check_path_values
>     raise AppMapInvalidConfigException(
> _appmap.configuration.AppMapInvalidConfigException: Malformed path value '.' in configuration file. Path entries must be module names.
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "/Users/ajp/.asdf/installs/python/3.10.13/lib/python3.10/site.py", line 617, in <module>
>     main()
>   File "/Users/ajp/.asdf/installs/python/3.10.13/lib/python3.10/site.py", line 610, in main
>     execsitecustomize()
>   File "/Users/ajp/.asdf/installs/python/3.10.13/lib/python3.10/site.py", line 549, in execsitecustomize
>     import sitecustomize
>   File "/Users/ajp/src/applandinc/worktrees/appmap-python/_appmap/test/data/fastapi/init/sitecustomize.py", line 1, in <module>
>     import appmap
>   File "/Users/ajp/src/applandinc/worktrees/appmap-python/appmap/__init__.py", line 9, in <module>
>     from _appmap import generation  # noqa: F401
>   File "/Users/ajp/src/applandinc/worktrees/appmap-python/_appmap/__init__.py", line 1, in <module>
>     from . import configuration
>   File "/Users/ajp/src/applandinc/worktrees/appmap-python/_appmap/configuration.py", line 450, in <module>
>     initialize()
>   File "/Users/ajp/src/applandinc/worktrees/appmap-python/_appmap/configuration.py", line 445, in initialize
>     Config().initialize()
>   File "/Users/ajp/src/applandinc/worktrees/appmap-python/_appmap/configuration.py", line 146, in __init__
>     self._load_config()
>   File "/Users/ajp/src/applandinc/worktrees/appmap-python/_appmap/configuration.py", line 230, in _load_config
>     sys.exit(1)
> SystemExit: 1
> ```
> 
> So, we need to change this. When an invalid `path` is detected`, we should just log a warning and ignore it. This needs to happen without raising an exception, and the agent should not exit.

Done

_appmap/configuration.py
```diff
@@ -269,6 +274,40 @@ def _load_functions(self):
...
+                has_separator = isinstance(path, str) and ('/' in path or '\\' in path)
+                logger.warning(
+                    f"Malformed path value '{path}' in configuration file. "
+                    "Path entries must be module names"
+                    f"{' not directory paths' if has_separator else ''}."
+                )
+                invalid_items.append(item)
+                continue
+
+        if len(invalid_items) > 0:
```
apotterri, line 300: What's going to happen if this eliminates all the path entries?

zermelo-wisen: I think, no package will be instrumented, unless there are some `dist` entries. We reach this state with at least one warning.

Thinking again, should we check if:
 1. `packages` is a yaml sequence and not a scalar
 2. `packages` is empty
 3. Every item has `path` XOR `dist` key

apotterri: No, let's not expand the scope of this beyond checking that `path`s are modules.

However, please add a test that has only invalid `path`s, and confirm that the resulting `Config` has a `package` with an empty list of `paths`.

zermelo-wisen: Done

apotterri (approved):

PR #304, opened 2024-05-13
feat: Search for config file

### LINKED ISSUE #295: The agent searches for appmap.yml
When looking for an existing `appmap.yml`, the agent should search from the current directory up to the repo root (or the root directory, if not in a repo), and use the first one it finds. `appmap_dir` should be resolved relative to the location of the discovered `appmap.yml`.

zermelo-wisen:
Fixes #295.

- In `Config._load_config`, if the config file is not found relative to the current folder (more precisely `Env.current.root_dir`) it is searched up to the repo root, or up to the file system root if there is no repo.

- `appmap_dir` is resolved relative to the location of config file and `Env.current.output_dir` is updated with this value, unless `APPMAP_OUTPUT_DIR` is set by tests.

- [Tests](https://github.com/getappmap/appmap-python/blob/4d0a29c761850b3320d67a0dfe86dca7b721aee2/_appmap/test/test_configuration.py#L263-L291) added to ensure:
-- Config file is found in a parent folder.
-- Config file is not found up until the repo root even if it exists in the parent directory of the repo root.
-- Config file is not found up until the file system root.

### DIFF
--- a/_appmap/test/test_util.py
+++ b/_appmap/test/test_util.py
@@ -2,7 +2,11 @@
 Test util functionality
 """
 
-from _appmap.utils import scenario_filename
+import os
+from pathlib import Path
+import uuid
+
+from _appmap.utils import locate_file_up, scenario_filename
 
 
 def test_scenario_filename__short():
@@ -13,3 +17,14 @@ def test_scenario_filename__short():
 def test_scenario_filename__special_character():
     """has a customizable suffix"""
     assert scenario_filename("foobar?=65") == "foobar_65"
+
+def test_locate_file_up(data_dir):
+    result = locate_file_up("appmap.yml", Path(data_dir) / "package1" / "package2")
+    assert result.parts[-3:] == ("_appmap", "test", "data")
+
+    result = locate_file_up("test_util.py", Path(data_dir) / "package1" / "package2")
+    assert result.parts[-2:] == ("_appmap", "test")
+
+    impossible_file_name = str(uuid.uuid4()) + ".yml"
+    result = locate_file_up(impossible_file_name, data_dir)
+    assert result is None
--- a/_appmap/utils.py
+++ b/_appmap/utils.py
@@ -1,5 +1,6 @@
 import inspect
 import os
+from pathlib import Path
 import re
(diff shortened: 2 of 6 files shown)


kgilpin:
Can you pls describe the solution in the PR description.

zermelo-wisen:
> Can you pls describe the solution in the PR description.

Done.

_appmap/test/test_util.py
```diff
@@ -13,3 +16,14 @@ def test_scenario_filename__short():
...
     assert scenario_filename("foobar?=65") == "foobar_65"
+
+def test_locate_file_up(data_dir):
+    result = locate_file_up("appmap.yml", os.path.join(data_dir, "package1", "package2"))
+    assert result.endswith(os.path.join("_appmap", "test", "data"))
+
+    result = locate_file_up("test_util.py", os.path.join(data_dir, "package1", "package2"))
+    assert result.endswith(os.path.join("_appmap", "test"))
+
+    impoosible_file_name = str(uuid.uuid4()) + ".yml"
```
apotterri, line 27:
```suggestion
    impossible_file_name = str(uuid.uuid4()) + ".yml"
```

_appmap/utils.py
```diff
@@ -222,3 +222,27 @@ def scenario_filename(name, separator="_"):
...
+    :param filename: The name of the file to locate.
+    :param start_dir: The directory to start the search from. Defaults to the current.
+    :param stop_idr: The directory to stıp the search. If None search is performed until
+                     the root of the file system.
+    :return: The path to the directory containing the file or None if the file cannot be found.
+    """
+    if start_dir is None:
+        start_dir = os.getcwd()
+
+    file_path = os.path.join(start_dir, filename)
```
apotterri, line 240: In general, it seems like it might be better to use `pathlib`. Was there a reason you chose `os.path` here?

apotterri (changes requested):
A couple of minor changes to make, but I also wonder about the choice to use `os.path` instead of `pathlib`.

apotterri (approved):

PR #305, opened 2024-05-14
disable recording by default

### LINKED ISSUE #299: The agent no longer records by default
Change the default for `APPMAP`, from `true` to `false`. Also, combine `APPMAP_RECORD_PYTEST` with `APPMAP_RECORD_UNITTEST` to `APPMAP_RECORD_TEST`, make sure `APPMAP_RECORD_REQUESTS` is disabled when it's set.

This is a breaking change, so the major version will need to update to v2.

### LINKED ISSUE #300: Request recording can't be enabled when running tests
It should be possible to enable request recording when running tests by setting `APPMAP_RECORD_REQUESTS=true`.  Currently, it's not: https://github.com/getappmap/appmap-python/blob/4ada3fe6c05169c2241adca438451aec16e9f1d5/_appmap/testing_framework.py#L120 https://github.com/getappmap/appmap-python/blob/4ada3fe6c05169c2241adca438451aec16e9f1d5/_appmap/env.py#L129-L132

### LINKED ISSUE #301: Test recording is controlled by a single env var
Currently, test recording is configured per test type, by `APPMAP_RECORD_PYTEST` and `APPMAP_RECORD_UNITTEST`. in v2, these should be combined into a single variable named `APPMAP_RECORD_TESTS`.

### LINKED ISSUE #307: Enabling process recording should disable all other recording methods
When the user sets `APPMAP_RECORD_PROCESS=true`, they're explicitly saying they'd like a process recording. This recording will show all the code that gets executed by the process, so there's not really any reason to enable the other types. This will also avoid errors.

apotterri:
The agent no longer records by default. Code must be run with `appmap-python`, or with `APPMAP=true` to create AppMap data files.

Additionally, `APPMAP_RECORD_PYTEST` and `APPMAP_RECORD_UNITTEST` have been replaced by `APPMAP_RECORD_TESTS`. Also, `APPMAP_RECORD_PROCESS` now disables other recording types.

Fixes #299.
Fixes #300.
Fixes #301.
Fixes #307.

### DIFF
--- a/appmap/__init__.py
+++ b/appmap/__init__.py
@@ -9,40 +9,40 @@ _enabled = os.environ.get("APPMAP", None)
 if _enabled is None or _enabled.upper() == "TRUE":
     if _enabled is not None:
         # Use setdefault so tests can manage _APPMAP as necessary
-        os.environ["_APPMAP"] = _enabled
-    from _appmap import generation  # noqa: F401
-    from _appmap.env import Env  # noqa: F401
-    from _appmap.importer import instrument_module  # noqa: F401
-    from _appmap.labels import labels  # noqa: F401
-    from _appmap.noappmap import decorator as noappmap  # noqa: F401
-    from _appmap.recording import Recording  # noqa: F401
+        os.environ.setdefault("_APPMAP", _enabled)
+        from _appmap import generation  # noqa: F401
+        from _appmap.env import Env  # noqa: F401
+        from _appmap.importer import instrument_module  # noqa: F401
+        from _appmap.labels import labels  # noqa: F401
+        from _appmap.noappmap import decorator as noappmap  # noqa: F401
+        from _appmap.recording import Recording  # noqa: F401
 
-    try:
-        from . import django  # noqa: F401
-    except ImportError:
-        pass
+        try:
+            from . import django  # noqa: F401
+        except ImportError:
+            pass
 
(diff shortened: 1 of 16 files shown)


appmap/__init__.py
```diff
@@ -9,40 +22,42 @@
...
+        # need to be imported here
+
+        # unittest is part of the standard library, so it should always be
+        # importable (and therefore doesn't need to be in a try .. except block)
+        from . import unittest  # noqa: F401
+
+        def enabled():
+            return Env.current.enabled
+else:
+    print(_DISABLED_BY_DEFAULT_MSG, file=sys.stderr)
```
dividedmind, line 63: Won't this cause the warning to be displayed even if the user explicitly sets `APPMAP=false`? 
This seems very annoying.

dividedmind (changes requested):
Looks mostly ok, though explicitly setting `APPMAP=false` (or anything else than `true`) should not cause a warning. (BTW, I noticed that with the previous version `APPMAP=false` doesn't seem to do anything at all — warning is still displayed and appmaps are still recorded.)

appmap/__init__.py
apotterri: Yeah, I noticed this was wrong right before I read your comment. I adjusted the indentation of the `else`, should work properly now.

PR #306, opened 2024-05-14
feat: append to a single log file

### LINKED ISSUE #296: The agent appends to a single log file, instead of creating a new one for each run
Currently, the agent creates a new log file, named with the current timestamp, each time it's run. Instead, it should only create a small number of files (maybe only 1), and append to it.

One approach could be to use RotatingFileHandler, with a low backup count.

zermelo-wisen:
Fixes #296.

I used RotatingFileHandler with max size of 5MB and backup count of 5.

File names will be like (after every 5MB):

appmap.log
appmap.1.log
appmap.2.log
appmap.3.log
appmap.4.log
appmap.5.log


_appmap/env.py
```diff
@@ -188,9 +188,11 @@ def _configure_logging(self):
             loggers["appmap"]["level"] = loggers["_appmap"]["level"] = log_level
             config_dict["handlers"] = {
                 "default": {
-                    "class": "logging.FileHandler",
+                    "class": "logging.handlers.RotatingFileHandler",
                     "formatter": "default",
-                    "filename": f"appmap-{now:%Y%m%d%H%M%S}-{os.getpid()}.log",
+                    "filename": "appmap.log",
+                    "maxBytes": 5 * 1024 * 1024,
+                    "backupCount": 5,
```
apotterri, line 195: Let's make `maxBytes` bigger, say `50 * 1024 * 1024`, and set `backupCount` to `1`.

apotterri (changes requested):
Now that we've switched to `RotatingFileHandler`, let's change the default for APPMAP_DISABLE_LOG_FILE to `false`: https://github.com/getappmap/appmap-python/blob/4ada3fe6c05169c2241adca438451aec16e9f1d5/_appmap/env.py#L157 .

apotterri (approved):

PR #308, opened 2024-05-15
fix SQLAlchemy deprecated execute param in test

### LINKED ISSUE #279: Support async ORM - sqlalchemy 2.0

zermelo-wisen:
Fixes #279.

- Fixed deprecated usages in tests. 
 -- Passing a string to `Conneciton.execute()` is deprecated in SQLAlchemy 1.4 and removed in 2.0: https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Connection.execute
-- Changed the tests with warnings issuing a DML with `Connection.execute()` to be explicit about commit behavior since auto-commit behavior is changed in 2.0: https://docs.sqlalchemy.org/en/20/changelog/migration_20.html
- Updated tox config to test with both 1.4 and 2.0 versions of SQLAlchemy.

### DIFF
--- a/_appmap/test/test_sqlalchemy.py
+++ b/_appmap/test/test_sqlalchemy.py
@@ -10,6 +10,7 @@ from sqlalchemy import (
     MetaData,
     String,
     Table,
+    text,
     create_engine,
 )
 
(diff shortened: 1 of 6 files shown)


zermelo-wisen:
> SQLAlchemy v1.4 is still getting updates, so we need to continue testing against it.
> 
> Can you update the tox config to include an environment that does that, in the same way we test old versions of Django and Flask, please?

Done.

apotterri (approved):

PR #309, opened 2024-05-17
fix: find a config in the repo root

### LINKED ISSUE #295: The agent searches for appmap.yml

apotterri:
Make sure a config that exists in the repo root will be found in any subdirectory, including an immediate child.

Fixes #295


_appmap/test/test_configuration.py
```diff
@@ -273,38 +274,49 @@ def test_missing_packages(self, tmpdir):
 class TestSearchConfig:
```
zermelo-wisen, line 274: I just thought about whether it would be a good idea to have a test to ensure that the search is stopped at the repository root.

Let's say we have an `a/b/c` folder hierarchy, where `a` has `appmap.yml`, `b` is the repository root, and `c` is the current folder. The test will ensure that the search starting from `c` should not find the config file since it's beyond the repository root.

zermelo-wisen (approved):

_appmap/test/test_configuration.py
apotterri: Done. Thanks.

PR #310, opened 2024-05-23
fix: handle non json serializable types

### LINKED ISSUE #283: Non-JSON serializable types throw an error when recorded with AppMap
Non-JSON serializable types can cause the application to panic.

```txt
FAILED tests/test.pt - TypeError: Object of type int64 is not JSON serializable
1 failed, 29 passed, 39 warnings in 10.53s
```

```txt
_appmap/generation.py:114: in default
    return json.JSONEncoder.default(self, o)
```

via https://appmap-group.slack.com/archives/C03GM7D2QRH/p1708366829827279

https://github.com/getappmap/appmap-python/issues/312

zermelo-wisen:
Fixes #283.
 
- Adds `numpy` as a dev dependency. v1.24.4 is the latest version supporting Python 3.8.
- Changes `tox.ini` to support different `numpy` versions in Python 3.8 and 3.{9,10,11,12}
- Adds tests for `AppMapEncoder` ensuring objects containing members with `numpy.int64` and `numpy.array` types can be serialized without errors.
- Modifies a test that generates an AppMap to cover the issue.

### DIFF
--- a/_appmap/test/test_generation.py
+++ b/_appmap/test/test_generation.py
@@ -1,4 +1,9 @@
+import json
 import pytest
+
+import numpy as np
+
+from _appmap.generation import AppMapEncoder
 
 
 @pytest.mark.appmap_enabled
@@ -48,3 +53,18 @@ class TestGeneration:
             return ret
 
         verify_example_appmap(check_comment, "instance_method")
+
+class TestAppMapEncoder:
+    def test_np_int64_type(self):
+        data = {
+            "value": np.int64(42),
+        }
+        json_str = json.dumps(data, cls=AppMapEncoder)
+        assert '{"value": "42"}' == json_str
+
+    def test_np_array_type(self):
+        data = {
+            "value": np.array([0, 1, 2, 3])
+        }
+        json_str = json.dumps(data, cls=AppMapEncoder)
+        assert '{"value": "[0 1 2 3]"}' == json_str
--- a/_appmap/test/data/pytest/expected/pytest.appmap.json
+++ b/_appmap/test/data/pytest/expected/pytest.appmap.json
@@ -8,95 +8,133 @@
       "name": "appmap",
       "url": "https://github.com/applandinc/appmap-python"
     },
+    "source_location": "test_simple.py:5",
+    "name": "hello world",
+    "feature": "Hello world",
     "app": "Simple",
     "recorder": {
       "name": "pytest",
       "type": "tests"
     },
-    "source_location": "test_simple.py:5",
-    "name": "hello world",
-    "feature": "Hello world",
     "test_status": "succeeded"
   },
   "events": [
     {
-      "defined_class": "simple.Simple",
-      "method_id": "hello_world",
-      "path": "simple.py",
-      "lineno": 8,
       "static": false,
       "receiver": {
-        "class": "simple.Simple",
         "kind": "req",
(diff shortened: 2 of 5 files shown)


_appmap/test/test_generation.py
```diff
@@ -48,3 +53,18 @@ def check_comment(self, to_dict):
             return ret
 
         verify_example_appmap(check_comment, "instance_method")
+
+class TestAppMapEncoder:
```
apotterri, line 57: This test is good. I think it would also be helpful, especially to understand the impact of the change, to update one of the tests that actually generates an AppMap. One of these two could be good: https://github.com/getappmap/appmap-python/blob/e4b8d78638e97527d30605643ddd2903aa063d55/_appmap/test/web_framework.py#L286, or https://github.com/getappmap/appmap-python/blob/master/_appmap/test/test_test_frameworks.py#L94

apotterri (changes requested):
One additional test to add. Also, the build is red, but maybe you already noticed that?

zermelo-wisen:
> One additional test to add. Also, the build is red, but maybe you already noticed that?

Yes, thank you. I just canceled my review request :)

_appmap/test/test_generation.py
zermelo-wisen: Actually, I tried to create exactly this as a separate AppMap generation test but wasn't able to produce the expected error - test passes without the fix as well. Returning int64 typed values from or accepting them as arguments to recorded functions seems not to produce the error. 

Here is a temp branch if you want to take a look what am I missing there: https://github.com/getappmap/appmap-python/tree/temp/test-recording-with-numpy-int64

apotterri: It's hard to tell why this isn't reproducing the expected behavior, because it's only checking the length of the event array.

The reason I picked those other two test cases was because they'd requiring updating the expected AppMap data.

Rather than entirely new test case, can you try updating one of those tests I suggested, please?

zermelo-wisen: It was not possible to produce a non JSON serializable AppMap by simply introducing a `numpy.int64` value as an event parameter or as a return value of an event, since all values are converted to strings with `event.py::display_string` while encoding their container `Event`.

While examining the `seaborn` project which had this [problem](https://github.com/getappmap/appmap-python/issues/312), it turned out that the issue occurs when `numpy.int64` is used as a key in a `dict` that is a parameter or a return value of a recorded function. These keys become the `name` attributes of properties while describing the value and the schema of the return value or the parameter. The `name` attributes are not checked or converted to strings in `event.py`, which leads to this error.

In addition to the direct tests for `AppMapEncoder`, I modified one of the existing tests you suggested to produce the AppMap that includes the recording of a function returning a `dict` with `numpy.int64` keys.

appmap.log
apotterri, line 1: Delete this, please.

apotterri (changes requested):
Thanks for separating out the testing updates into a separate commit, this makes it easier to understand what you did. However, please combine them into a single commit after you remove `appmap.log`. 

I like to use `fixup!` commits to accomplish this: https://git-scm.com/docs/git-commit/2.32.0#Documentation/git-commit.txt---fixupamendrewordltcommitgt .

appmap.log
zermelo-wisen: Oh sorry...

apotterri (approved):

PR #327, opened 2024-05-29
fix: appmap.Recording is available even when APPMAP=false

### LINKED ISSUE #326: `appmap.Recording` is available even when `APPMAP=false`
Code that imports `appmap.Recording` should run correctly even when `not Env.current.enabled`.

It may be possible to implement this by simply moving the import from `_appmap` outside the current `if..then` that's keeping it from being imported.

zermelo-wisen:
Fixes #326.

- Adds `NoopRecording` class with the same interface as `Recording`.
- Exports `NoopRecording` as `Recording` when `APPMAP` is not set to `true` to prevent "ImportError: cannot import name 'Recording' from 'appmap' ..." in client code.
- Extends [`smoketest.sh`](https://github.com/getappmap/appmap-python/blob/325a58b7981f7e994626b29193ba89e4814bc8d4/ci/smoketest.sh#L31-L48) to ensure that client code using `appmap.Recording` does not fail when `APPMAP=false`.

### DIFF
--- a/ci/smoketest.sh
+++ b/ci/smoketest.sh
@@ -1,5 +1,24 @@
 #!/usr/bin/env bash
 
+test_recording_when_appmap_not_true()
+{
+  cat <<EOF > test_client.py
+from appmap import Recording
+
+with Recording():
+  print("Hello from appmap library client")
+EOF
+
+  python test_client.py
+
+  if [[ $? -eq 0 ]]; then
+    echo 'Script executed successfully'
+  else
+    echo 'Script execution failed'
+    exit 1
+  fi
+}
+
 set -ex
 pip -q install -U pip pytest "flask>=2,<3" python-decouple
 pip -q install /dist/appmap-*-py3-none-any.whl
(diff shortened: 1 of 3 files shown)


ci/smoketest.sh
```diff
@@ -27,3 +27,22 @@ else
   find $PWD
   exit 1
 fi
+
+# Ensure that client code using appmap.Recording does not fail when APPMAP=false
```
apotterri, line 31: This can get moved up before https://github.com/getappmap/appmap-python/pull/327/files#diff-ba2ff002750238ee03b3467fd4400804141c8aa2606bdd4263f0a3636bb120c1R12 to avoid flipping `APPMAP` on and off.

This file is starting to get kind of complicated. Please move this test into a separate function. Thanks!

apotterri (changes requested):
Thanks for this, looks good. One minor change and it'll be ready to go.

apotterri (approved):
Thanks, good to go.

PR #328, opened 2024-05-29
fix: optionally limit number of events collected

### LINKED ISSUE #329: The APPMAP_MAX_EVENTS environment variables limits the number of events added to a recording.
if `APPMAP_MAX_EVENTS` is set, it's interpreted as an `int` that limits the number of events that can be added to a Recording. Once the limit is reached, the exception `_appmap.recorder.AppMapTooManyEvents` is raised, and subsequent events are ignored.

apotterri:
If the environment variable APPMAP_MAX_EVENTS is set, use it to limit the number of events added to a Recorder. This setting is a suggestion, rather than a hard limit, but the Recorder should get close.

Fixes #329 .

In the interest of getting this out ASAP, it doesn't have any additional tests. It shouldn't have any impact when it's unset, which is shown by the fact that all the existing tests pass.

### DIFF
--- a/_appmap/recorder.py
+++ b/_appmap/recorder.py
@@ -10,6 +10,17 @@ logger = Env.current.getLogger(__name__)
 # pylint: disable=global-statement
 _default_recorder = None
 
+# Allow the user to suggest a limit on the number of events that should be added to a Recorder.
+# Depending on how exceptions get processed by the framework, there may be some more added, but it
+# shouldn't be an enormous number.
+_MAX_EVENTS = Env.current.get("APPMAP_MAX_EVENTS")
+if _MAX_EVENTS is not None:
+    _MAX_EVENTS = int(_MAX_EVENTS)
+
+
+class AppMapTooManyEvents(RuntimeError):
+    """Thrown when a recorder has more than APPMAP_MAX_EVENTS"""
+
 
 class Recorder(ABC):
     """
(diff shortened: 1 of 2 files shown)


_appmap/recorder.py
```diff
@@ -18,6 +29,8 @@ class Recorder(ABC):
     Note that the abstract methods have implementations for use by subclasses.
     """
 
+    _aborting = False
```
kgilpin, line 32: Is there any chance that this will cause unintended consequences? I guess it’s so there is only one exception per recorder. 

Other recorders (eg different tests that may be running concurrently or subsequently) will continue?

dustinbyrne: From what I've seen it's only possible to have one active `Recorder` at a time, though I don't see why this couldn't be a member variable

apotterri: The variable is shared by all instances of Recorder, so the global Recorder and request recorders all see the same value. Once one of them has a lot of events, they'll all stop adding events. 

How concurrent tests behave will depend on how they're isolated. I think we've got bigger problems than this if they share state.

Subsequent tests reset `_aborting`.

apotterri: > From what I've seen it's only possible to have one active `Recorder` at a time, though I don't see why this couldn't be a member variable

You definitely can, because you can have a global recorder and a per-thread recorder. Sharing this ensures they'll all abort once things get out of hand.

This is working for me in Django now, and since it's disabled by default, end-users won't see it. If you all would like to use soon, we can leave it as-is, and revisit it later.

kgilpin (approved):

PR #337, opened 2024-06-05
Add APPMAP_MAX_TIME

apotterri:
Add `APPMAP_MAX_TIME` which allows the user to specify, in seconds, the maximum time a recording session should be allowed.

In addition, disable rendering of the schema associated with values (parameter, return types) when `APPMAP_DISPLAY_PARAMS=false`.

Finally, move the implementation of `__reduce_ex__` from `wrapt.wrappers.FunctionWrapper` up to `ObjectProxy`. Our implementation, which is good enough for generating AppMaps, works for any instance of an ObjectProxy.

### DIFF
--- a/_appmap/event.py
+++ b/_appmap/event.py
@@ -111,7 +111,9 @@ def describe_value(name, val, max_depth=5):
         "object_id": id(val),
         "value": display_string(val),
     }
-    ret.update(_describe_schema(name, val, 0, max_depth))
+    if Env.current.display_params:
+        ret.update(_describe_schema(name, val, 0, max_depth))
+
     if any(_is_list_or_dict(type(val))):
         ret["size"] = len(val)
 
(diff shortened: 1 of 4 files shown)


kgilpin:
I'm curious if pytest and unittest also have a test timeout setting?

How does test timeout setting differ in practice from a max events setting?

apotterri:
`pytest` has a timeout plugin: https://pypi.org/project/pytest-timeout/ . As far as I can tell, `unittest` doesn't support a global timeout, it requires you to use something like https://pypi.org/project/timeout-decorator/ .

In both of those cases, the test is going to get aborted, and no AppMap is going to get generated. Using `APPMAP_MAX_TIME` means that we'll still get an AppMap, with a test status showing us why it failed.

Setting `APPMAP_MAX_EVENTS` wasn't sufficient for dealing with memory issues, because there are other things that can cause high memory utilization. This PR has a fix for one of them (disabling the generation of `properties` for hash-like objects), but I expect that are others as well.

kgilpin:
Cool, thanks for the explanation.

dustinbyrne (approved):

PR #338, opened 2024-06-07
fix: request recording in unittest setUp method

### LINKED ISSUE #332: request recordings generated when testing Django
Running the Django tests (which are based on `unittest`) for Django v4.1 generates request recordings in addition to recordings for the test cases. The agent is trying to [disable request recording](https://github.com/getappmap/appmap-python/blob/47d8cc32adfbce52d7f935dd68229eb7454e66c1/_appmap/testing_framework.py#L120), but this doesn't appear to be sufficient.

zermelo-wisen:
Fixes #332

- Hooks `unittest` `TestCase._callSetUp` method to disable request recording in order to prevent the creation of a request recording when a request is made inside `setUp` method of a `unittest` test.
- Adds a `unittest` test to `django` fixture project which makes a request inside `setUp` method. This test makes the `test_django.py::TestDjangoApp::test_enabled` pass with the fix and fail without the fix.
- This edge case can be observed in this test in `django` project:
   `$ APPMAP=TRUE ./runtests.py auth_tests.test_views.ChangelistTests.test_user_change_email`
A request made inside `ChangelistTests.setUp` method leads to the creation of requests recording AppMap without the fix.

### DIFF
--- a/_appmap/test/test_django.py
+++ b/_appmap/test/test_django.py
@@ -201,11 +201,13 @@ class TestDjangoApp:
         # To really check middleware reset, the tests must run in order,
         # so disable randomly.
         result = pytester.runpytest("-svv", "-p", "no:randomly")
-        result.assert_outcomes(passed=4, failed=0, errors=0)
+        result.assert_outcomes(passed=5, failed=0, errors=0)
         # Look for the http_server_request event in test_app's appmap. If
         # middleware reset is broken, it won't be there.
         appmap_file = pytester.path / "tmp" / "appmap" / "pytest" / "test_request.appmap.json"
-        assert not os.path.exists(pytester.path / "tmp" / "appmap" / "requests")
+        assert not os.path.exists(
+            pytester.path / "tmp" / "appmap" / "requests"
+        ), "django tests generated request recordings"
 
         events = json.loads(appmap_file.read_text())["events"]
         assert "http_server_request" in events[0]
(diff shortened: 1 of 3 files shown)


_appmap/test/test_django.py
```diff
@@ -213,7 +213,7 @@ def test_enabled(self, pytester):
     def test_disabled(self, pytester, monkeypatch):
         monkeypatch.setenv("_APPMAP", "false")
         result = pytester.runpytest("-svv", "-p", "no:randomly", "-k", "test_request")
-        result.assert_outcomes(passed=1, failed=0, errors=0)
+        result.assert_outcomes(passed=2, failed=0, errors=0)
```
apotterri, line 216: This also needs to check that no request recordings were created.

apotterri (changes requested):

_appmap/test/test_django.py
zermelo-wisen: Doesn't the following line assert that as well?

https://github.com/getappmap/appmap-python/blob/f90ca457a33bdb9cd4eb808b3da27e9accb3e2d4/_appmap/test/test_django.py#L217

apotterri: Well, it does, in the case that `APPMAP=false` (i.e. recording is completely disabled). The test I asked for ensures that request recordings aren't generated when other AppMaps are.

apotterri (approved):
Looks good. Thanks!

PR #340, opened 2024-06-17
fix: appmap breaks vscode python extension starting a REPL

### LINKED ISSUE #339: Installing appmap breaks Microsoft's VS Code python extension support for starting a REPL
If you install `appmap` into the virtual environment selected for a project, trying to start a REPL (with the command `python.execInREPL`, which is bound to `shift+enter` by default) will fail.

Turning on logging for `vscode-python` shows the stack trace:
```
2024-06-14 09:22:29.976 [error] Error: Error processing line 1 of /opt/anaconda3/envs/python-test/lib/python3.9/site-packages/appmap.pth:


	at s (/Users/ajp/.vscode/extensions/ms-python.python-2024.8.1/out/client/extension.js:2:224147)
	at Socket.<anonymous> (/Users/ajp/.vscode/extensions/ms-python.python-2024.8.1/out/client/extension.js:2:224443)
	at Socket.emit (node:events:514:28)
	at addChunk (node:internal/streams/readable:376:12)
	at readableAddChunk (node:internal/streams/readable:349:9)
	at Readable.push (node:internal/streams/readable:286:10)
	at Pipe.onStreamRead (node:internal/stream_base_commons:190:23)
```
in the "Extension Host" output.

zermelo-wisen:
Fixes #339.

- Tolerates the inability to create the `appmap.log` file for logging, since a file system read-only error can occur in the [VSCode Python Extension, Virtual Environment, REPL, Shift Enter] scenario when attempting to create the log file.
- Mounts a read-only `appmap.log` file to Docker and adds a test to the smoke test suite for running with a non-modifiable `appmap.log`.

### DIFF
--- a/_appmap/env.py
+++ b/_appmap/env.py
@@ -37,6 +37,7 @@ class Env(metaclass=SingletonMeta):
                 else:
                     self._env.pop(k, None)
 
+        self.log_file_creation_failed = False
         self._configure_logging()
         enabled = self._env.get("_APPMAP", "false")
         self._enabled = enabled is None or enabled.lower() != "false"
@@ -133,6 +134,21 @@ class Env(metaclass=SingletonMeta):
     def getLogger(self, name) -> trace_logger.TraceLogger:
         return cast(trace_logger.TraceLogger, logging.getLogger(name))
 
+    def determine_log_file(self):
+        log_file = "appmap.log"
+
+        # Try creating the log file in the current directory
+        try:
+            with open(log_file, 'a', encoding='UTF8'):
+                pass
+        except IOError:
+            # The circumstances in which creation is going to fail
+            # are also those in which the user doesn't care whether
+            # there's a log file (e.g. when starting a REPL).
+            return None
+        return log_file
+
+
     def _configure_logging(self):
         trace_logger.install()
 
@@ -179,13 +195,19 @@ class Env(metaclass=SingletonMeta):
             log_level = self.get("APPMAP_LOG_LEVEL", "info").upper()
             loggers = config_dict["loggers"]
             loggers["appmap"]["level"] = loggers["_appmap"]["level"] = log_level
+
+            log_file = self.determine_log_file()
+            # Use NullHandler if log_file is None to avoid complicating the configuration
+            # with the absence of the "default" handler.
             config_dict["handlers"] = {
                 "default": {
                     "class": "logging.handlers.RotatingFileHandler",
                     "formatter": "default",
-                    "filename": "appmap.log",
+                    "filename": log_file,
                     "maxBytes": 50 * 1024 * 1024,
                     "backupCount": 1,
+                } if log_file is not None else {
+                    "class": "logging.NullHandler"
                 },
                 "stderr": {
                     "class": "logging.StreamHandler",
@@ -194,6 +216,7 @@ class Env(metaclass=SingletonMeta):
                     "stream": "ext://sys.stderr",
                 },
             }
+            self.log_file_creation_failed = log_file is None
 
         if log_config is not None:
             name, level = log_config.split("=", 2)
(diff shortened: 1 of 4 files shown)


_appmap/env.py
```diff
@@ -133,6 +134,29 @@ def display_params(self):
...
         return cast(trace_logger.TraceLogger, logging.getLogger(name))
 
+    def determine_log_file(self):
+        log_file = "appmap.log"
+
+        # Try creating the log file in the current directory
+        try:
+            with open(log_file, 'a', encoding='UTF8'):
+                pass
+        except IOError:
```
apotterri, line 144: I don't think we want to create the log file in the user's home directory, or as a temp file.

Instead, let's just write a message to stderr return `None`. The circumstances in which creation is going to fail are also those in which the user doesn't care whether there's a log file (e.g. when starting a REPL).

_appmap/env.py
```diff
@@ -183,7 +208,7 @@ def _configure_logging(self):
                 "default": {
```
apotterri, line 208: If `determine_log_file` returns `None`, we don't need to add a handler here. Writing warnings to stderr is enough.

_appmap/test/test_env.py
```diff
@@ -11,3 +12,17 @@ def test_disable_temporarily():
     except RuntimeError:
         ...
     assert env.enables("requests")
+
+def test_determine_log_file(monkeypatch):
+    env = Env({"_APPMAP": "true"})
```
apotterri, line 17: Did you check to see if `APPMAP` was really `"true"` when you were investigating #339? I don't think it should have been.

_appmap/test/test_env.py
```diff
@@ -11,3 +12,17 @@ def test_disable_temporarily():
...
+
```
apotterri, line 18: I think it's important to add an end-to-end test that actually shows that the behavior described in #339 is fixed. It would be ideal to add this to the smoketest, but that's going to be complicated by the fact that processes in docker run as root.

Can you give some thought to how we might implement such a test, please?

apotterri (changes requested):
Good start, a bit more to do.

_appmap/env.py
```diff
@@ -213,3 +236,7 @@ def initialize(**kwargs):
     Env.reset(**kwargs)
     logger = logging.getLogger(__name__)
     logger.info("appmap enabled: %s", Env.current.enabled)
+    if Env.current.log_file_creation_failed:
+        # Writing to stderr makes the REPL fail in vscode-python.
+        # https://github.com/microsoft/vscode-python/blob/c71c85ebf3749d5fac76899feefb21ee321a4b5b/src/client/common/process/rawProcessApis.ts#L268-L269
+        logger.info("appmap.log cannot be created")
```
apotterri, line 242: This needs to be logged at `warning` instead of `info` (because that's the level set for the `stderr` handler).

apotterri (changes requested):
Needs one more minor change, then this will be good to go.

The smoke test is starting to get pretty complicated. At some point, we could maybe simplify it by using something like [bats](https://hub.docker.com/r/bats/bats)

_appmap/env.py
zermelo-wisen: While debugging with VSCode Python extension sources, I found out that writing to `stderr` also breaks REPL. That's why I'm using `logger.info` here.

It can be seen from these lines:
https://github.com/microsoft/vscode-python/blob/c71c85ebf3749d5fac76899feefb21ee321a4b5b/src/client/common/process/rawProcessApis.ts#L268-L269

apotterri: Well, that seems like a poor choice on their part, but I guess we're stuck with it. :|

apotterri (approved):

_appmap/env.py
zermelo-wisen: Yes, and unfortunately, there could be some configuration combinations with appmap environment variables that would lead to `logger.warning` calls during the initialization steps, triggering the same problem.

zermelo-wisen: One ugly hack would be to suppress them, or convert them to logger.info calls if we detect that `appmap.log` is not writable.

PR #346, opened 2024-07-02
fix: bad appmaps in django oscar

zermelo-wisen:
- Modifies Django `Middleware` to work when there is no `headers` attribute in the response, but there is a `_headers` attribute instead. This issue was observed while running Django Oscar tests with AppMap Python on a specific version (git commit hash: 9574ce2a56f3d99836c2e20785c116da12af6c42) of Django Oscar.
- Adds a test to ensure that `Middleware.__call__` works when `get_response` returns a response with a `_headers` attribute instead of `headers`.

### DIFF
--- a/appmap/django.py
+++ b/appmap/django.py
@@ -224,12 +224,20 @@ class Middleware(AppmapMiddleware):
             self.on_exception(rec, start, call_event_id, sys.exc_info())
             raise
 
+        # response.headers attribute is missing here while running some
+        # django-oscar project tests. There is a response._header attribute
+        # instead.
+        #   django-oscar commit hash: 9574ce2a56f3d99836c2e20785c116da12af6c42
+        #   example test: tests/functional/dashboard/test_dashboard.py
+        #                 TestDashboardIndexForAnonUser::test_is_not_available
+        headers = getattr(response, 'headers', getattr(response, '_headers', None))
+
         self.after_request_hook(
             request.path_info,
             request.method,
             request.build_absolute_uri(),
             response.status_code,
-            response.headers,
+            headers,
             start,
             call_event_id,
         )
--- a/_appmap/test/test_django.py
+++ b/_appmap/test/test_django.py
@@ -186,6 +186,27 @@ def test_deeply_nested_routes(client, events):
     )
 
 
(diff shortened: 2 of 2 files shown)


_appmap/recording.py
```diff
@@ -81,6 +81,23 @@ def __exit__(self, exc_type, exc_value, tb):
         return False
 
 
+def check_call_return_stack_order(events):
```
apotterri, line 84: We don't want to do this in the agent. AppMaps can be quite large, and this will have a significant performance impact. Detecting the problem during fingerprinting is good enough

appmap/django.py
```diff
@@ -224,12 +224,20 @@ def __call__(self, request):
             self.on_exception(rec, start, call_event_id, sys.exc_info())
             raise
 
+        # response.headers attribute is missing here while running some
+        # django-oscar project tests. There is a response._header attribute
+        # instead.
+        #   django-oscar commit hash: 9574ce2a56f3d99836c2e20785c116da12af6c42
+        #   example test: tests/functional/dashboard/test_dashboard.py
+        #                 TestDashboardIndexForAnonUser::test_is_not_available
+        headers = getattr(response, 'headers', getattr(response, '_headers', None))
```
apotterri, line 233: Why does trying to use `headers` instead of `_headers` cause the event-ordering problem?

apotterri (changes requested):

appmap/django.py
zermelo-wisen: While running a `django-oscar` test, here "AttributeError: 'HttpResponseRedirect' object has no attribute 'headers'" is raised in our code. `pytest` framework catches it and marks the test as a failed test. We are able to write an AppMap for the failed test after the failed test finishes, as usual. However, our code which creates the http response return event does not get chance to run because of this exception. This produces a bad AppMap with an http request call event missing a matching http response return event.

_appmap/recording.py
zermelo-wisen: I removed this commit.

(My reasoning was that it would have an overhead but this overhead would always go with the main overhead of creating large AppMaps and it would be small compared to the large AppMap creation itself.)

PR #349, opened 2024-07-04
fix: bad appmaps in pytest project tests

zermelo-wisen:
- Catching `Exception` in the `instrument.py::call_instrumented` function to create exception return event, does not catch the `Skipped` exception raised from the `pytest.skip` function when the `pytest` library is instrumented because the `Skipped` exception is derived from `BaseException` but not from `Exception`. This leads to the bad AppMap due to the missing return event for the instrumented `pytest.skip` function.

- This fixes the problem by catching `BaseException`, the root class of the exception hierarchy, in `call_instrumented` function to be able to create the exception return event even if the instrumented function raises an exception not derived from the `Exception` class.

- Adds a test that records a method raising a `BaseException` which fails with producing a bad AppMap (unbalanced call order, missing exception return event) without the fix, and passes with the fix.

### DIFF
--- a/_appmap/instrument.py
+++ b/_appmap/instrument.py
@@ -101,7 +101,12 @@ def call_instrumented(f, instance, args, kwargs):
         return ret
     except AppMapLimitExceeded:
         raise
-    except Exception:  # noqa: E722
+    # Some applications make use of exceptions that aren't descended from Exception. For example,
+    # pytest's OutcomeException, used to indicate the outcome of a test case, is a child of
+    # BaseException.
+    #
+    # We need to catch *any* exception raised, to ensure that we add the appropriate ExceptionEvent.
+    except BaseException:  # noqa: E722
         elapsed_time = time.time() - start_time
         Recorder.add_event(
             event.ExceptionEvent(
--- a/_appmap/test/data/example_class.py
+++ b/_appmap/test/data/example_class.py
@@ -5,6 +5,7 @@ and called. Used for testing appmap instrumentation.
 
 import time
 from functools import lru_cache, wraps
+from typing import NoReturn
 
 import appmap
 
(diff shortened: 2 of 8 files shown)


_appmap/recording.py
```diff
@@ -80,6 +80,22 @@ def __exit__(self, exc_type, exc_value, tb):
             self.exit_hook(self)
         return False
 
+def check_call_return_stack_order(events):
```
apotterri, line 83: As with the django PR, this doesn't belong here. Checking it in the test case is ok, though, so moving it to test_events.py is fine.

_appmap/instrument.py
```diff
@@ -101,7 +101,7 @@ def call_instrumented(f, instance, args, kwargs):
         return ret
     except AppMapLimitExceeded:
         raise
-    except Exception:  # noqa: E722
+    except BaseException:  # noqa: E722
```
apotterri, line 104: Looking at the standard library's exception hierarchy, it's not clear to me why this change fixes the problem. Is there some other exception that inherits from `BaseException` that's getting raised?

apotterri (changes requested):

_appmap/instrument.py
zermelo-wisen: In `pytest` project a `Skipped` exception is raised in `skip` function here: https://github.com/pytest-dev/pytest/blob/main/src/_pytest/outcomes.py#L160. However, `Skipped` does not derive from `Exception`, it is (indirectly) derived from `BaseException`. We were not catching it in `_appmap/instrument.py::call_instrumented` and we were not able to create the return leg of the call, the exception return event. Missing this event resulted in a "bad" AppMap.

apotterri: Thanks for this explanation. Can you add a skipped test to https://github.com/getappmap/appmap-python/blob/master/_appmap/test/data/pytest/test_simple.py, please, to show that you've fixed the problem that was originally reported? Thanks!

apotterri (approved):

PR #354, opened 2024-07-22
try to avoid recording tests

### LINKED ISSUE #351: Tests are not included in AppMaps
Including test functions in recordings isn't helpful. Their presence is just noise, and confuses users (and Navie).

The default configuration generated by appmap-python should not include directories that primarily contain tests. Because there's no enforced standard for naming of test directories, appmap-python will exclude directories that match the regex `.*test.*`. 

In addition, test framework integration will disable recording of test functions, in case some tests have been instrumented (e.g. because they're in a module with a name that doesn't match `.*test.*`). This behavior is controlled by a property named `record_test_cases` in `appmap.yml`, which defaults to `false`.

apotterri:
When creating the default config, ignore directories with names that
match the regex .*test.*. This should avoid instrumenting the majority
of test functions. For those that do still get instrumented, have the
test framework integration disable the wrapt wrapped on it, thereby
disabling recording.

Also, some issues I found with the way the f{set,get,del} functions of properties were handled. And, adds an environment variable to disable instrumenting properties altogether.

Fixes #351


_appmap/importer.py
```diff
@@ -194,27 +196,34 @@ def do_import(cls, *args, **kwargs):
...
+                    filterableFn = FilterableFn(filterable, prop_name, fn, fn, auxtype)
+                    if getattr(fn, "_appmap_instrumented", None):
                         continue
-                    filterableFn = FilterableFn(filterable, fn, fn, auxtype)
-                    new_fn = cls.instrument_function(fn.__name__, filterableFn, selected_functions)
+                    new_fn = cls.instrument_function(prop_name, filterableFn, selected_functions)
                     if new_fn != fn:
                         new_fn = wrapt.FunctionWrapper(fn, new_fn)
-                        setattr(new_fn, "_appmap_wrapped", True)
+                        # Set _appmap_wrapped on the FunctionWrapper, not on the wrapped function
```
dividedmind, line 224: _appmap_instrumented

vendor/_appmap/wrapt/wrappers.py
```diff
@@ -524,6 +525,7 @@ def __init__(self, wrapped, instance, wrapper, enabled=None,
         object.__setattr__(self, '_self_binding', binding)
         object.__setattr__(self, '_self_parent', parent)
         object.__setattr__(self, '_bfws', list())
+        object.__setattr__(self, "_appmap_instrumented", False)
```
dividedmind, line 531: I'm not sure how I feel about modifying a vendored lib. I'm worried it'll be too easy to miss it when we update or if we ever decide to switch to non-vendored.

dividedmind (approved):

_appmap/importer.py
apotterri: Good catch, thanks.

vendor/_appmap/wrapt/wrappers.py
apotterri: This isn't the first change I've had to make to `wrapt`, so it's already the case that we have to merge from upstream, rather than just overwriting. Those other changes will also keep us from switching to non-vendored....

PR #385, opened 2026-04-14
faster regex::compile

theonlychant:
Adds a lightweight regex compilation booster and a small ML utilities package: __init__.py implements cached `compile` plus `match`/`search`/`findall`/`sub` and will use the `regex` package if available (falling back to stdlib `re`), and strongly_convex.py provides a pure‑Python averaged SGD implementation for λ‑strongly convex objectives with a demo in example_usage.py; I updated django.py and flask.py to import the cached shim (`from appmap import regex_boost as re`) to reduce repeated compilation overhead, and added BOOSTERS_THEOREMS.md containing concise theorem statements and proof sketches for the boosters. The changes are backward‑compatible (no new required dependencies), and you can sanity‑check with `python3 appmap/ml/example_usage.py` and by exercising existing Django/Flask paths in the test suite.


dividedmind:
I don't understand the problem this is trying to solve. The pull request also includes a lot of unrelated (what looks like junk) code and documentation. If there is a performance problem you're seeing, please describe it in more detail, if possible profile to pinpoint the cause and show how any changes improve things.

theonlychant:
yea the junk code part was from my old commits in 2023 for some reason it didn't delete when i made this PR sorry thanks for reviewing though

